Issue #2, opened 2020-07-24
What about products in L?


dwincort:
The current definition of `L` is:
```haskell
data L :: (* -> *) -> (* -> *) -> (* -> *) where
  Zero :: L f g s
  Scale :: s -> L Par1 Par1 s
  JoinL :: V h => h (L f g s) -> L (h :.: f) g s
  ForkL :: V h => h (L f g s) -> L f (h :.: g) s
```
This n-ary representation of join and fork feels nice, but it's also limiting.  For instance, how do you represent a matrix with 7 rows?  You'll need a representable functor `h` that has cardinality 7.  There's no way to build it up from smaller components using the machinery here.  In other words, there's no way to join a 5x2 matrix with a 2x2 matrix to get a 7x2 matrix.

The obvious way to fix this is to support constructors like
```haskell
  JoinL' :: V h => h (L f g s) -> h' (L f g s) -> L (h :*: h') g s
  ForkL' :: V h => h (L f g s) -> h' (L f g s) -> L f (h :*: h') s
```
but I'm not sure what the repercussions are.  At the very least, operations become much more cumbersome because there are so many constructors.

conal:
Thanks. I agree and have been adding binary nonuniform join & fork, essentially as you’ve suggested. Working out very nicely. Almost ready to commit.

bolt12:
Since this might be develop into more structured discussions as we go on on the design space I suggest every commit to be a separate Pull Request so people can better keep track and discuss things before merge.

conal:
Thanks for the suggestion. I just made pull request \#9. I wasn’t sure whether to merge before or after discussion. I chose the former, because I want to fix issue \#4 as well as a separate pull request. Can you still discuss after I merge?

I don’t have much experience with collaborative GitHub work, and I appreciate feedback & advice.

bolt12:
Yes you can still discuss after merge! It was a good idea to create a new "experimental" branch.

conal:
> But I don’t see any need for `ins` and `exs`. They do not make sense to me. Maybe I’m wrong here but `inl`/`inr` and `exl` and `exr` suffice.

@bolt12: Given my previous comment in this thread, do you now see the sense/meaning and purpose/value of `ins` and `exs` and the insufficiency of `(inl,inr)` and `(exl,exr)` in the presence of n-ary coproducts?

[...]

PR #9, opened 2020-07-24
Conal.dev

conal:
Adding binary, nonuniform join & fork, addressing issue \#2.

The dev branch and pull request allows others to easily comment on lines and regions.

### DIFF
--- a/src/LinAlg.hs
+++ b/src/LinAlg.hs
@@ -3,13 +3,26 @@
 module LinAlg where
 
 import qualified Prelude as P
-import Prelude hiding ((+),sum,(*))
+import Prelude hiding ((+),sum,(*),unzip)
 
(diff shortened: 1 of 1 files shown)


bolt12:
I'm not very fond of adding 2 new constructors, isn't it possible to have "n-ary" Join/Fork that support an arbitrary number of columns/rows ?

conal:
> I’m not very fond of adding 2 new constructors, isn’t it possible to have “n-ary” Join/Fork that support an arbitrary number of columns/rows ?

I’m not very fond of that choice either. Currently there are binary nonuniform and n-ary uniform fork/join. Are you suggesting n-ary *nonuniform*? If so, I don’t know how the types would work out. Maybe something like [`HList`](https://hackage.haskell.org/package/HList).

Issue #8, opened 2020-07-24
Pattern purpose


bolt12:
What's the purpose of the following patterns:

```haskell
pattern Fork :: V h => h (L f g s) -> L f (h :.: g) s
pattern Fork ms <- (unforkL -> ms) where Fork = ForkL
{-# complete Fork #-}

pattern Join :: V h => h (L f g s) -> L (h :.: f) g s
pattern Join ms <- (unjoinL -> ms) where Join = JoinL
{-# complete Join #-}
```

?

I'm not very familiar with the concept and never have used it in Haskell.

conal:
I wanted to capture the fact that fork and join are both *isomorphisms* and make it convenient to view linear maps as forks or joins whenever the types allow them to be.

This way, individual operations such as composition don’t need to explicitly reshape representations. For instance, note the following clauses for linear map addition:

``` haskell
  (f :|# g) + (h :| k) = (f + h) :| (g + k)
  (f :&# g) + (h :& k) = (f + h) :& (g + k)
  ForkL ms  + Fork ms' = Fork (liftR2 (+) ms ms')
  JoinL ms  + Join ms' = Join (liftR2 (+) ms ms')
```

The type synonyms cause the second argument in each clause to get reshaped as needed to match the shape of the corresponding first argument, resulting in code that closely resembles basic properties of abelian categories, including linear algebra.

conal:
An explanation I wrote in my notes and then forgot:

The fork and join operations (whether binary nonuniform or n-ary uniform) are *isomorphisms*. For instance, not only can we fork two linear maps with common domain to get a linear map to pairs, but *every* linear map to pairs is a fork. Similarly for join, and similarly for n-ary uniform fork and join. Pattern matching is about invertibility/injectivity, allowing one to write in terms of forward operations, implying the inverse computation without having to say it explicitly. Pattern synonyms give us the notational convenience of pattern matching without the rigidity of directly mirroring the representation. For instance, the `Fork` pattern matches a literal `ForkL`, but also sometimes a `JoinL`.

PR #11, opened 2020-07-24
Fixes #4

### LINKED ISSUE #4: idL is not the identity of Semiring multiplication over L
If `L` represents matrices, then `idL` is the matrix of all ones, not the identity matrix.  This means that the `Semiring` instance isn't quite right.

conal:


src/LinAlg.hs
```diff
@@ -137,31 +143,40 @@ instance Additive s => Additive (L f g s) where
   Zero + m = m
   m + Zero = m
   Scale s + Scale s' = Scale (s + s') -- distributivity
-  (f :|# g) + (h :| k) = (f + h) :|# (g + k)
-  (f :&# g) + (h :& k) = (f + h) :&# (g + k)
-  ForkL ms + Fork ms' = ForkL (liftR2 (+) ms ms')
-  JoinL ms + Join ms' = JoinL (liftR2 (+) ms ms')
-  ForkL ms + Fork ms' = ForkL (liftR2 (+) ms ms')
+  (f :|# g) + (h :| k) = (f + h) :| (g + k)
```
bolt12, line 146: @conal I'm not sure about this because the pattern synonyms confuse me. But this might only work when adding two matrices that have exactly the same permutation of Join/Fork. I ran into that problem in "TYM" and I mention there what I did to solve it. It might be more complex with this data structure if this is really have the same issue, but it's possible

bolt12: I see now your answer in #8. Does this means that what I mentioned is not a problem?

conal: > I see now your answer in \#8. Does this means that what I mentioned is not a problem?

Yes, exactly. The synonyms give us views of the second argument reshaped as necessary to match the outer structure of the first argument.

If you wouldn’t mind pointing me to the section of *TYM* that discusses this problem and your solution, I’d enjoy comparing solutions.

conal: For instance, consider

``` haskell
  (f :|# g) + (h :| k) = (f + h) :| (g + k)
```

If I replace the “semantic pattern” `h :| k` with a “syntactic pattern” `h :|# k`, GHC rightly complains about partiality:

    Pattern match(es) are non-exhaustive
    In an equation for ‘+’:
        Patterns not matched:
            (_ :|# _) (_ :&# _)
            (_ :|# _) (ForkL _)

With the semantic pattern, those two cases and `_ :|# _` are all handled in one (perhaps deceptively) simple clause.

bolt12: @conal 

| If you wouldn’t mind pointing me to the section of TYM that discusses this problem and your solution, I’d enjoy comparing solutions.

Section 4, last paragraph. Your solution is similar to mine but I do not use pattern synonyms :) 

It is a clever solution that I couldn't come up with! We're always learning!

src/LinAlg.hs
```diff
@@ -137,31 +143,40 @@ instance Additive s => Additive (L f g s) where
...
+matrix' :: (V2 f g, V2 h k) => k (h (L f g s)) -> L (h :.: f) (k :.: g) s
+matrix' = Fork . fmap Join
 
 matrix :: (V f, V g) => g (f s) -> L (f :.: Par1) (g :.: Par1) s
-matrix = Fork . fmap (Join . fmap Scale)
+matrix = matrix' . (fmap.fmap) Scale
+
+diagR :: (Representable h, Eq (Rep h), Additive a) => h a -> h (h a)
+diagR as =
+  tabulate (\ i -> (tabulate (\ j -> if i == j then as `index` i else zero)))
```
dwincort, line 159: I think you overthought this one.  You can get away with the simpler:
```haskell
diagR :: (Representable h, Eq (Rep h), Semiring a) => a -> h (h a)
diagR a =
  tabulate (\ i -> (tabulate (\ j -> if i == j then a else zero)))

idL :: (V f, Semiring s) => L (f :.: Par1) (f :.: Par1) s
idL = rowMajor (diagR one)
```

conal: Thanks. I expected to want the more general `diagR`, but I'll restrict as you suggested and re-generalize later if motivated.

dwincort: Hmm, I'm not sure I see why the generality of yours is particularly desired.  Or, put another way, if you're going to be general, I'm not sure why you wouldn't have gone fully general, as in:
```haskell
diagR :: (Representable h, Eq (Rep h), Additive a) => h (h a) -> h (h a)
diagR as =
  tabulate (\ i -> (tabulate (\ j -> if i == j then as `index` i `index` j else zero)))

idL = matrix (diagR (pureRep (pureRep one)))
```

dwincort: Oh I think I see.  In your version of the function, the input `h a` is, in fact, a representation of the _diagonal itself_, and in the identity, the diagonal is precisely `pureRep one`.  I didn't appreciate that earlier.

PR #20, opened 2020-07-26
Khatri

bolt12:
- Adds matrix transposition
- Adds Khatri Rao projections
- Adds Khatri Rao product

I'm just a bit concerned about the first projection type signature: fstM :: (Semiring s, HasScaleV m, V k, Bang k) => L (m :.: k) (m :.: Par1) s it should only be m instead of (m :.: Par1). Although it is still "correct", it makes the type signatures a bit convoluted... I don't think there's any way around that.

I'm very happy with how this is turning out! The ability to perform "n-ary" joins/forks allowed me to implement the khatri-rao projections without recurring to dependent types or type classes or manual indice manipulation as I do in my LAoP library!

### DIFF
--- a/src/LinAlg.hs
+++ b/src/LinAlg.hs
@@ -1,5 +1,3 @@
-{-# LANGUAGE UndecidableInstances #-} -- see below
-
 -- | Linear algebra after Fortran
 
 module LinAlg where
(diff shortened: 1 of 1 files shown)


bolt12:
@conal I created a branch out of yours so I think everything is cool now!

conal:
Oops. I guess these issues are being discussed at PR #21. Is there a way we might have been able to see changes for PRs #18, #20, and #21 separately?

bolt12:
@conal  I don't really know.. I already made a branch out of other branches and the previous commits seem to prevail.

Issue #10, opened 2020-07-24
Khatri-Rao Product


bolt12:
I'm curious to see if the Khatri Rao product can be easily typeable. I think it should be something like this:

khatri :: V h => h (L f j s) :.: g (L f j s) -> L f (h :.: g) s

Because the khatri-rao of two matrices A = 3x4 and B = 3x5, should yield a matrix C = 3x(4x5) = 3x20, so the Khatri-Rao product can be seen like matrix pairing.

But looking at it, I think this type signature can also be employed by ForkL since it is more general (might be wrong here, once again I'm still trying to figure out how this works). If this is the case I'm not sure if we should allow "n-ary" Joins and Forks

bolt12:
See #19 . The type signature ended needing to be: `khatri :: (...) => L f (m :.: Par1) s -> L f k s -> L f (m :.: k) s` which I'm not sure we can improve upon, but it works! :smiley:

Issue #6, opened 2020-07-24
Kronecker Product


dwincort:
What exactly _is_ the kronecker product?  As in, what is its denotational meaning over linear maps?  Wikipedia answers _how_ to solve it but not really what it means.

bolt12:
I suggest you to read "Matrices as arrows Why categories of matrices matter". There you can find a very interesting relation of some of linear algebra's operators (such as the Kronecker product) as extensions of the Relational Algebra ones.

Kronecker Product is defined on top of the Khatri-Rao product. The Khatri-Rao product can be seen as matrix pairing, if you think in terms of relations `(a, b) (R ▽ S) c == a R b && a S c`, where `▽` is the Khatri Rao Product operator. The Khatri-Rao can be expressed using intersection and cartesian projection `λ(y, x) = y and ρ(y, x) = x: R ▽ S = (λº . R) ∩ (ρº . S)`. Note that Categorically this is not a true product.

So now you can derive the Kronecker from the Khatri Rao: `R × S = (λ . R) ▽ (ρ . S)`, expressing the "running in parallel''
of two logic expressions, by unpacking pairs on both sides: `(a, b)(R × S)(c, d)`, meaning the same as `a R c ∧ b S d`
 
I hope this connection to Relational algebra and logic makes it clear to you the meaning of the Kronecker Product. If this sounds to "magical" to you (as it seemed to me at first :smile: ) I recommend you to experiment some of these operators with my [laop library](https://github.com/bolt12/laop])

conal:
@bolt12: Do you have a sense of what `khatri` *means* in terms of linear maps/functions or just as a manipulation of the matrix representation? Same question for Kronecker product, as in issue \#6.

bolt12:
@conal 

> Do you have a sense of what khatri means in terms of linear maps/functions

In terms of normal functions I'd say that Khatri is the same as pairing, so something like a function of type `a -> b -> (a, b)`.

> manipulation of the matrix representation?

Manipulation of the matrix representation in order to implement the khatri rao is quite tricky. You'll have a better time by implementing the khatri rao projections `λ` and `p`, which can be defined inductively or "imperatively" once you know that `λ ▽ p = id`, and then `a ▽ b = (λº . a) * (ρº . b)`. So, `λ` is the matrix of type `(m * k) -> k` and `p` is of type `(m * k) -> k`. 

An example of a `λ` matrix:

```
(3 * 3) -> 3

111 000 000
000 111 000
000 000 111
```

[...]

PR #21, opened 2020-07-26
Eq

bolt12:
- Adds Eq instance to `L` type

It also gets rid of a unnecessary constrain on the pattern synonyms.

### DIFF
--- a/src/LinAlg.hs
+++ b/src/LinAlg.hs
@@ -1,5 +1,3 @@
-{-# LANGUAGE UndecidableInstances #-} -- see below
-
 -- | Linear algebra after Fortran
 
 module LinAlg where
@@ -8,7 +6,7 @@ import qualified Prelude as P
 import Prelude hiding ((+),sum,(*),unzip)
 
 import GHC.Generics (Par1(..), (:*:)(..), (:.:)(..))
-import Control.Arrow ((***))
+import qualified Control.Arrow as A
 import Data.Distributive
 import Data.Functor.Rep
 
@@ -59,11 +57,25 @@ data L :: (* -> *) -> (* -> *) -> (* -> *) where
   JoinL :: V h => h (L f g s) -> L (h :.: f) g s
   ForkL :: V h => h (L f g s) -> L f (h :.: g) s
 
+-- Scalable vectors
+class V a => HasScaleV a where
+  scaleV :: Semiring s => s -> L a a s
+
+type HasScaleV2 a b = (HasScaleV a, HasScaleV b)
+
+instance HasScaleV Par1 where scaleV = Scale
+
+instance (HasScaleV a, HasScaleV b) => HasScaleV (a :*: b) where
+  scaleV s = scaleV s *** scaleV s
+
+instance (HasScaleV a, HasScaleV b, Representable b) => HasScaleV (b :.: a) where
+  scaleV s = cross (pureRep (scaleV s))
+
 unjoin2 :: Additive s => L (f :*: g) h s -> L f h s :* L g h s
 unjoin2 Zero = (zero,zero)
 unjoin2 (p :|# q) = (p,q)
 unjoin2 ((unjoin2 -> (p,q)) :&# (unjoin2 -> (r,s))) = (p :& r, q :& s)
-unjoin2 (ForkL ms) = (ForkL *** ForkL) (unzip (unjoin2 <$> ms))
+unjoin2 (ForkL ms) = (ForkL A.*** ForkL) (unzip (unjoin2 <$> ms))
 
 -- unjoin2 ((p :| q) :&# (r :| s)) = (p :&# r, q :#& s)
 
@@ -83,7 +95,7 @@ unfork2 :: Additive s => L f (h :*: k) s -> L f h s :* L f k s
 unfork2 Zero = (zero,zero)
 unfork2 (p :&# q) = (p,q)
 unfork2 ((unfork2 -> (p,q)) :|# (unfork2 -> (r,s))) = (p :|# r, q :|# s)
-unfork2 (JoinL ms) = (JoinL *** JoinL) (unzip (unfork2 <$> ms))
+unfork2 (JoinL ms) = (JoinL A.*** JoinL) (unzip (unfork2 <$> ms))
 
 -- unfork2 ((p :& q) :|# (r :& s)) = (p :|# r, q :|# s)
 
@@ -160,8 +172,8 @@ diagR :: (Representable h, Eq (Rep h), Additive a) => h a -> h (h a)
 diagR as =
   tabulate (\ i -> (tabulate (\ j -> if i == j then as `index` i else zero)))
 
-idL :: (V f, Semiring s) => L (f :.: Par1) (f :.: Par1) s
-idL = rowMajor (diagR (pureRep one))
+idL :: (HasScaleV a, Semiring s) => L a a s
+idL = scaleV one
 
 infixr 9 .@
 (.@) :: Semiring s => L g h s -> L f g s -> L f h s
@@ -173,12 +185,59 @@ m         .@ (p :|# q) = (m .@ p) :|# (m .@ q)     -- binary coproduct law
 (r :|# s) .@ (p :&# q) = (r .@ p) + (s .@ q)       -- biproduct law
 ForkL ms' .@ m         = ForkL (fmap (.@ m) ms')   -- n-ary product law
 m'        .@ JoinL ms  = JoinL (fmap (m' .@) ms)   -- n-ary coproduct law
-JoinL ms' .@ ForkL ms  = sum (liftR2 (.@) ms' ms)  -- biproduct law
+JoinL ms' .@ ForkL ms  = sum (ms' .@* ms)          -- biproduct law
 
-instance (V f, Semiring s) => Semiring (L (f :.: Par1) (f :.: Par1) s) where
+(.@*) :: (Representable p, Semiring s) => p (L g h s) -> p (L f g s) -> p (L f h s)
+(.@*) = liftR2 (.@)
+
+instance (HasScaleV f, Semiring s) => Semiring (L f f s) where
   one = idL
   (*) = (.@)
 
--- Illegal nested constraint ‘Eq (Rep f)’
--- (Use UndecidableInstances to permit this)
+-- Injections
 
+inl :: (HasScaleV a, Semiring s) => L a (a :*: b) s 
+inl = idL :& zero
+
+inr :: (HasScaleV b, Semiring s) => L b (a :*: b) s 
+inr = zero :& idL
+
+-- Projections
+
+exl :: (HasScaleV a, Semiring s) => L (a :*: b) a s 
+exl = idL :| zero
+
+exr :: (HasScaleV b, Semiring s) => L (a :*: b) b s 
+exr = zero :| idL
+
+-- Injections
+ins :: (HasScaleV2 a c, Semiring s) => c (L a (c :.: a) s)
+ins = unjoinL idL
+
+-- Projections
+exs :: (HasScaleV2 a c, Semiring s) => c (L (c :.: a) a s)
+exs = unforkL idL
+
+-- Binary biproduct bifunctor
+(***) :: L a c s -> L b d s -> L (a :*: b) (c :*: d) s
+f *** g = (f :|# Zero) :&# (Zero :|# g)
+
+-- N-ary biproduct bifunctor
+cross :: (V a, HasScaleV2 b c, Semiring s) => c (L a b s) -> L (c :.: a) (c :.: b) s
+cross fs = JoinL (ins .@* fs)
+
+#if 0
+-- Equivalently,
+
+f *** g = (f :&# Zero) :|# (Zero :&# g)
+
+(***) :: (HasScaleV2 a b, Semiring s) => L a c s -> L b d s -> L (a :*: b) (c :*: d) s
+f *** g = (f .@ exl) :&# (g .@ exr)
+
+(***) :: (HasScaleV2 c d, Semiring s) => L a c s -> L b d s -> L (a :*: b) (c :*: d) s
+f *** g = (inl .@ f) :|# (inr .@ g)
+
+-- Equivalently,
+cross :: (V a, HasScaleV2 a c, Semiring s) => c (L a b s) -> L (c :.: a) (c :.: b) s
+cross fs = ForkL (fs .@* exs)
+#endif
--- a/src/LinAlg.hs
+++ b/src/LinAlg.hs
@@ -185,10 +185,10 @@ m         .@ (p :|# q) = (m .@ p) :|# (m .@ q)     -- binary coproduct law
 (r :|# s) .@ (p :&# q) = (r .@ p) + (s .@ q)       -- biproduct law
 ForkL ms' .@ m         = ForkL (fmap (.@ m) ms')   -- n-ary product law
 m'        .@ JoinL ms  = JoinL (fmap (m' .@) ms)   -- n-ary coproduct law
-JoinL ms' .@ ForkL ms  = sum (ms' .@* ms)          -- biproduct law
+JoinL ms' .@ ForkL ms  = sum (ms' .^ ms)          -- biproduct law
 
-(.@*) :: (Representable p, Semiring s) => p (L g h s) -> p (L f g s) -> p (L f h s)
-(.@*) = liftR2 (.@)
+(.^) :: (Representable p, Semiring s) => p (L g h s) -> p (L f g s) -> p (L f h s)
+(.^) = liftR2 (.@)
 
 instance (HasScaleV f, Semiring s) => Semiring (L f f s) where
   one = idL
@@ -224,7 +224,7 @@ f *** g = (f :|# Zero) :&# (Zero :|# g)
 
 -- N-ary biproduct bifunctor
 cross :: (V a, HasScaleV2 b c, Semiring s) => c (L a b s) -> L (c :.: a) (c :.: b) s
-cross fs = JoinL (ins .@* fs)
+cross fs = JoinL (ins .^ fs)
 
 #if 0
 -- Equivalently,
@@ -239,5 +239,5 @@ f *** g = (inl .@ f) :|# (inr .@ g)
 
 -- Equivalently,
 cross :: (V a, HasScaleV2 a c, Semiring s) => c (L a b s) -> L (c :.: a) (c :.: b) s
-cross fs = ForkL (fs .@* exs)
+cross fs = ForkL (fs .^ exs)
 #endif
--- a/src/LinAlg.hs
+++ b/src/LinAlg.hs
@@ -185,7 +185,7 @@ m         .@ (p :|# q) = (m .@ p) :|# (m .@ q)     -- binary coproduct law
 (r :|# s) .@ (p :&# q) = (r .@ p) + (s .@ q)       -- biproduct law
 ForkL ms' .@ m         = ForkL (fmap (.@ m) ms')   -- n-ary product law
 m'        .@ JoinL ms  = JoinL (fmap (m' .@) ms)   -- n-ary coproduct law
-JoinL ms' .@ ForkL ms  = sum (ms' .^ ms)          -- biproduct law
+JoinL ms' .@ ForkL ms  = sum (ms' .^ ms)           -- biproduct law
 
 (.^) :: (Representable p, Semiring s) => p (L g h s) -> p (L f g s) -> p (L f h s)
 (.^) = liftR2 (.@)
--- a/src/LinAlg.hs
+++ b/src/LinAlg.hs
@@ -57,6 +57,16 @@ data L :: (* -> *) -> (* -> *) -> (* -> *) where
   JoinL :: V h => h (L f g s) -> L (h :.: f) g s
   ForkL :: V h => h (L f g s) -> L f (h :.: g) s
 
+instance Eq s => Eq (L f g s) where
+  Zero == Zero            = True
+  Zero == _               = False
+  _ == Zero               = False
+  (Scale s) == (Scale s') = s == s'
+  (f :|# g) == (h :| k)   = (f == h) && (g == k)
+  (f :&# g) == (h :& k)   = (f == h) && (g == k)
+  ForkL ms  == Fork ms'   = and $ liftR2 (==) ms ms'
+  JoinL ms  == Join ms'   = and $ liftR2 (==) ms ms'
+
 -- Scalable vectors
 class V a => HasScaleV a where
   scaleV :: Semiring s => s -> L a a s
@@ -71,8 +81,8 @@ instance (HasScaleV a, HasScaleV b) => HasScaleV (a :*: b) where
 instance (HasScaleV a, HasScaleV b, Representable b) => HasScaleV (b :.: a) where
   scaleV s = cross (pureRep (scaleV s))
 
-unjoin2 :: Additive s => L (f :*: g) h s -> L f h s :* L g h s
-unjoin2 Zero = (zero,zero)
+unjoin2 :: L (f :*: g) h s -> L f h s :* L g h s
+unjoin2 Zero = (Zero,Zero)
 unjoin2 (p :|# q) = (p,q)
 unjoin2 ((unjoin2 -> (p,q)) :&# (unjoin2 -> (r,s))) = (p :& r, q :& s)
 unjoin2 (ForkL ms) = (ForkL A.*** ForkL) (unzip (unjoin2 <$> ms))
@@ -91,19 +101,19 @@ unjoin2 (ForkL ms) = (ForkL A.*** ForkL) (unzip (unjoin2 <$> ms))
 (ForkL *** ForkL) (unzip (unjoin <$> ms)) :: L f (k :.: h) s :* L g (k :.: h) s
 #endif
 
-unfork2 :: Additive s => L f (h :*: k) s -> L f h s :* L f k s
-unfork2 Zero = (zero,zero)
+unfork2 :: L f (h :*: k) s -> L f h s :* L f k s
+unfork2 Zero = (Zero,Zero)
 unfork2 (p :&# q) = (p,q)
 unfork2 ((unfork2 -> (p,q)) :|# (unfork2 -> (r,s))) = (p :|# r, q :|# s)
 unfork2 (JoinL ms) = (JoinL A.*** JoinL) (unzip (unfork2 <$> ms))
 
 -- unfork2 ((p :& q) :|# (r :& s)) = (p :|# r, q :|# s)
 
-pattern (:&) :: Additive s => L f h s -> L f k s -> L f (h :*: k) s
+pattern (:&) :: L f h s -> L f k s -> L f (h :*: k) s
 pattern u :& v <- (unfork2 -> (u,v)) where (:&) = (:&#)
 {-# complete (:&) #-}
 
-pattern (:|) :: Additive s => L f h s -> L g h s -> L (f :*: g) h s
+pattern (:|) :: L f h s -> L g h s -> L (f :*: g) h s
 pattern u :| v <- (unjoin2 -> (u,v)) where (:|) = (:|#)
 {-# complete (:|) #-}
 
--- a/src/LinAlg.hs
+++ b/src/LinAlg.hs
@@ -170,10 +170,30 @@ rowMajor = rowMajor' . (fmap.fmap) Scale
 
 diagR :: (Representable h, Eq (Rep h), Additive a) => h a -> h (h a)
 diagR as =
-  tabulate (\ i -> (tabulate (\ j -> if i == j then as `index` i else zero)))
+  tabulate (\ i -> tabulate (\ j -> if i == j then as `index` i else zero))
 
 idL :: (HasScaleV a, Semiring s) => L a a s
 idL = scaleV one
+
+class V a => Bang a where
+  bang :: Semiring s => L a Par1 s
+
+instance Bang Par1 where bang = Scale one
+
+instance (Bang a, Bang b) => Bang (a :*: b) where
+  bang = bang :|# bang
+
+instance (Bang a, Bang b) => Bang (a :.: b) where
+  bang = JoinL (pureRep bang)
+
+-- Matrix transpose
+tr :: L c r s -> L r c s
+tr Zero = Zero
+tr (Scale s) = Scale s
+tr (a :|# b) = tr a :&# tr b
+tr (a :&# b) = tr a :|# tr b
+tr (ForkL m) = JoinL (fmap tr m)
+tr (JoinL m) = ForkL (fmap tr m)
 
 infixr 9 .@
 (.@) :: Semiring s => L g h s -> L f g s -> L f h s
@@ -196,18 +216,18 @@ instance (HasScaleV f, Semiring s) => Semiring (L f f s) where
 
 -- Injections
 
-inl :: (HasScaleV a, Semiring s) => L a (a :*: b) s 
+inl :: (HasScaleV a, Semiring s) => L a (a :*: b) s
 inl = idL :& zero
 
-inr :: (HasScaleV b, Semiring s) => L b (a :*: b) s 
+inr :: (HasScaleV b, Semiring s) => L b (a :*: b) s
 inr = zero :& idL
 
 -- Projections
 
-exl :: (HasScaleV a, Semiring s) => L (a :*: b) a s 
+exl :: (HasScaleV a, Semiring s) => L (a :*: b) a s
 exl = idL :| zero
 
-exr :: (HasScaleV b, Semiring s) => L (a :*: b) b s 
+exr :: (HasScaleV b, Semiring s) => L (a :*: b) b s
 exr = zero :| idL
 
 -- Injections
@@ -221,6 +241,9 @@ exs = unforkL idL
 -- Binary biproduct bifunctor
 (***) :: L a c s -> L b d s -> L (a :*: b) (c :*: d) s
 f *** g = (f :|# Zero) :&# (Zero :|# g)
+
+(+++) :: L a c s -> L b d s -> L (a :*: b) (c :*: d) s
+f +++ g = (f :|# Zero) :&# (Zero :|# g)
 
 -- N-ary biproduct bifunctor
 cross :: (V a, HasScaleV2 b c, Semiring s) => c (L a b s) -> L (c :.: a) (c :.: b) s
@@ -241,3 +264,18 @@ f *** g = (inl .@ f) :|# (inr .@ g)
 cross :: (V a, HasScaleV2 a c, Semiring s) => c (L a b s) -> L (c :.: a) (c :.: b) s
 cross fs = ForkL (fs .^ exs)
 #endif
+
+-- Khatri-Rao projections
+
+fstM :: (Semiring s, HasScaleV m, V k, Bang k) => L (m :.: k) (m :.: Par1) s
+fstM = cross (pureRep bang)
+
+sndM :: (Semiring s, HasScaleV k, V m) => L (m :.: k) k s
+sndM = JoinL (pureRep idL)
+
+-- Khatri Rao product (Matrix pairing)
+
+khatri :: (Semiring s, Semiring (L f (m :.: k) s), Bang k, HasScaleV2 m k)
+       => L f (m :.: Par1) s -> L f k s -> L f (m :.: k) s
+khatri a b = (tr fstM .@ a) * (tr sndM .@ b)
+


bolt12:
Can I add the following type family (as described in "TYM") to efficiently convert a TypeLit to a balanced tree?

```haskell
type family FromNat (n :: Nat) :: (Type -> Type) where
  FromNat 1 = Par1
  FromNat n = FromNat' (Mod n 2 == 0) (FromNat (Div n 2))

type family FromNat' (b :: Bool) (m :: Type -> Type) :: (Type -> Type) where
  FromNat' 'True m  = m :*: m
  FromNat' 'False m = Par1 :*: (m :*: m)
```

This might make our life easier to test for larger matrices

src/LinAlg.hs
```diff
@@ -59,11 +57,35 @@ data L :: (* -> *) -> (* -> *) -> (* -> *) where
   JoinL :: V h => h (L f g s) -> L (h :.: f) g s
   ForkL :: V h => h (L f g s) -> L f (h :.: g) s
 
-unjoin2 :: Additive s => L (f :*: g) h s -> L f h s :* L g h s
-unjoin2 Zero = (zero,zero)
+instance Eq s => Eq (L f g s) where
+  Zero == Zero            = True
+  Zero == _               = False
+  _ == Zero               = False
+  (Scale s) == (Scale s') = s == s'
```
dwincort, line 64: This `Eq` instance defines structural equality but not semantic equality.  For example, `Zero` and `Scale 0` are semantically equivalent (when `Zero` is at the right type), but they will be marked as not equal.  What about changing these first few cases to:
```haskell
Zero == Zero        = True
Zero == m           = m == Zero
Scale s == Scale s' = s == s'
Scale s == Zero     = s == 0
```

src/LinAlg.hs
```diff
@@ -158,10 +180,30 @@ rowMajor = rowMajor' . (fmap.fmap) Scale
...
 diagR :: (Representable h, Eq (Rep h), Additive a) => h a -> h (h a)
 diagR as =
-  tabulate (\ i -> (tabulate (\ j -> if i == j then as `index` i else zero)))
+  tabulate (\ i -> tabulate (\ j -> if i == j then as `index` i else zero))
+
+idL :: (HasScaleV a, Semiring s) => L a a s
+idL = scaleV one
+
+class V a => Bang a where
+  bang :: Semiring s => L a Par1 s
```
dwincort, line 189: What exactly does `bang` do?

src/LinAlg.hs
```diff
@@ -173,12 +215,77 @@ m         .@ (p :|# q) = (m .@ p) :|# (m .@ q)     -- binary coproduct law
...
+-- Projections
+exs :: (HasScaleV2 a c, Semiring s) => c (L (c :.: a) a s)
+exs = unforkL idL
+
+-- Binary biproduct bifunctor
+(***) :: L a c s -> L b d s -> L (a :*: b) (c :*: d) s
+f *** g = (f :|# Zero) :&# (Zero :|# g)
+
+(+++) :: L a c s -> L b d s -> L (a :*: b) (c :*: d) s
+f +++ g = (f :|# Zero) :&# (Zero :|# g)
```
dwincort, line 256: This looks like some duplicated code.

src/LinAlg.hs
```diff
@@ -173,12 +215,77 @@ m         .@ (p :|# q) = (m .@ p) :|# (m .@ q)     -- binary coproduct law
...
+fstM = cross (pureRep bang)
+
+sndM :: (Semiring s, HasScaleV k, V m) => L (m :.: k) k s
+sndM = JoinL (pureRep idL)
+
+-- Khatri Rao product (Matrix pairing)
+
+khatri :: (Semiring s, Semiring (L f (m :.: k) s), Bang k, HasScaleV2 m k)
+       => L f (m :.: Par1) s -> L f k s -> L f (m :.: k) s
+khatri a b = (tr fstM .@ a) * (tr sndM .@ b)
```
dwincort, line 290: The `Semiring (L f (m :.: k) s)` constraint feels a little strange to me.  Why not omit it and use `.@` in place of `*`?

src/LinAlg.hs
```diff
@@ -173,12 +215,77 @@ m         .@ (p :|# q) = (m .@ p) :|# (m .@ q)     -- binary coproduct law
...
+cross fs = ForkL (fs .^ exs)
+#endif
+
+-- Khatri-Rao projections
+
+fstM :: (Semiring s, HasScaleV m, V k, Bang k) => L (m :.: k) (m :.: Par1) s
+fstM = cross (pureRep bang)
+
+sndM :: (Semiring s, HasScaleV k, V m) => L (m :.: k) k s
+sndM = JoinL (pureRep idL)
```
dwincort, line 284: I know very little about khatri-rao, but it strikes me as a bit strange how different these two projections are.  It makes me think something somewhere is overly complicated.

src/LinAlg.hs
```diff
@@ -158,10 +180,30 @@ rowMajor = rowMajor' . (fmap.fmap) Scale
...
+instance Bang Par1 where bang = Scale one
+
+instance (Bang a, Bang b) => Bang (a :*: b) where
+  bang = bang :|# bang
+
+instance (Bang a, Bang b) => Bang (a :.: b) where
+  bang = JoinL (pureRep bang)
+
+-- Matrix transpose
+tr :: L c r s -> L r c s
```
conal, line 200: Do you have a meaning in mind for `tr` (in terms of the meaning of `L`)?

conal: I think this question has quite a lovely answer, btw.

src/LinAlg.hs
conal: I think we’re still missing a precise specification of `khatri`, although discussions are started in issues \#6 and \#10. Likewise for the corresponding the corresponding projections. I don’t know how to interpret or evaluate an implementation/answer without a specification/question.

conal: These signatures look suspicious to me also. In my experience, clarifying the specification helps tremendously in choosing and evaluating APIs.

conal:
> Can I add the following type family (as described in “TYM”) to efficiently convert a TypeLit to a balanced tree?

I expect that this question will lead to an interesting and useful discussion of its own, including motivation and philosophy. I’d like to see it moved to a new issue, where it can be found and read more easily.

src/LinAlg.hs
bolt12: Yes, this is a good addition! Thank you!

src/LinAlg.hs
bolt12: Bang (!) is the special case of the **top** matrix (**1**), which is the unit of matrix multiplication `m x n`. Bang has `n = 1`.

This terminology draws an analogy with the homonym function in the algebra of programming: the "bang function'', in general of type `A -> 1` and usually denoted by symbol `!`, is the polymorphic constant function yielding the unique value which inhabits the singleton type 1.

It's useful to define the Khatri-Rao product projections!

src/LinAlg.hs
bolt12: Yes I just added because I thought the `Arrow` naming convention was being followed and I just wanted to emphasize the biproduct here so `(+++) == (***)`. Also I'd suggest renaming `(:|#)` and `(:&#)` to maybe `(|||)` and `(===)`, respectively, following the same convention.

src/LinAlg.hs
bolt12: Well `tr` is the beginning of the `tr`anspose word so I chose that abbreviation :stuck_out_tongue: 

I think there's a Haskell extension that let's us use unary postfix operators which might look cool in this case and rename `tr` to `º`.

src/LinAlg.hs
conal: What's the *top* matrix?

src/LinAlg.hs
bolt12: @dwincort `.@` is different from `*` so I don't think that's really an option. But I agree it is indeed strange but that's what GHC asks for ahah.

@conal the standard definition of the Khatri-Rao product is presented as a (column-wise) variant of the Kronecker Product.
The projections need to reflect each other, this means: `khatri fstM sndM == idL`. I guess I can not really give a good answer to your questions regarding "what it means", I only know that it means `matrix pairing` and we can use it to apply an uncurried binary function to two matrices, for example. [This](https://pdfs.semanticscholar.org/ccf5/27fa9179081223bffe8067edd81948644fc0.pdf) and [this](https://repositorium.sdum.uminho.pt/handle/1822/22894) might give you more details.

src/LinAlg.hs
dwincort: Between this and the comment about writing `tr` as a postfix unicode character, I'll point out that Haskell _does_ have support for unicode if we want to use it.  For instance, using `∘` instead of `.@` for composition is pretty nice.

Also, I haven't thought this through, but might it make sense to have, say `(+++)` defined the other way, as in:
```haskell
(+++) :: L a c s -> L b d s -> L (a :*: b) (c :*: d) s
f +++ g = (f :&# Zero) :|# (Zero :&# g)
```
Semantically, they're the same, but I wonder if these two different operations could impact performance.

src/LinAlg.hs
conal: This doesn’t seem right to me, as there is not only one linear map from vectors to scalars, i.e., `Par1` is not terminal. Shouldn’t `bang` instead have type `L a U1 s`? Note that `Rep Par1 = ()`, and `Rep U1 = Void`.

On the other hand, `L a Par1 s =~ a =~ L Par1 a s`, the very useful isomorphisms between any vector space `a` over scalars `s` and the space of linear maps from `a` to `s` and from `s` to `a`.

src/LinAlg.hs
dwincort: > @dwincort `.@` is different from `*` so I don't think that's really an option. But I agree it is indeed strange but that's what GHC asks for ahah.

But `*` is defined as `.@` for linear maps, and we're very clearly composing two linear maps here.

src/LinAlg.hs
conal: Is the modified definition correct, i.e., consistent with linear map equality? Does anyone want to propose what linear map equality is?

bolt12: I'm not sure, but it looks good to me. I mean `Zero :: L Par1 Par1 s` can only be the same as `Scale 0`, right?

src/LinAlg.hs
conal: > I just added because I thought the `Arrow` naming convention was being followed and I just wanted to emphasize the biproduct here so `(+++) == (***)`.

Doesn’t “`(+++)`” emphasize the *coproduct* (as `(***)` emphasizes the product) rather than the biproduct? I don’t know how to emphasize the biproduct nature. In your LAOP work, the objects are indices, while in this representation the objects are the corresponding tries (whose logs are those indices), so wherever you say type sums, we instead have functor products. For this reason, I tend to use the name “`(***)`” in place of “`(+++)`”.

On the other hand, there are probably many situations in which the natural generalization from biproduct categories to bicartesian categories (having products and coproducts but not necessarily equal) would be `(+++)`, so I’m not so certain about eliminating the name “`(+++)`”. I suspect this issue will come up again.

> Also I’d suggest renaming `(:|#)` and `(:&#)` to maybe `(|||)` and `(===)`, respectively, following the same convention.

I assume you meant “`(***)`” and not “`(===)`” here. I chose the names to reflect this convention, and originally called them “`(:|||)`” and “`(:***)`” before deciding to give my finger and eyes a break. For pattern matching, we still have to use operator names that start with colons for either the syntactic or semantic constructors (literal constructors or pattern synonyms).

src/LinAlg.hs
bolt12: @conal The top matrix is the matrix full of ones, is a concept that comes from Relational Algebra, which is the largest Boolean matrix.

`bang` must be a column vector so I do not think that it should be `U1`. I just gave it this name because my background comes from Algebra of Programming and a lot of the concepts transpose to linear algebra. Once we use quantitative matrices instead of boolean matrices the **top** matrix stops being the **top** matrix as known in Rel, obviously

src/LinAlg.hs
bolt12: Yes, you can write it the that way! Hopefully it shouldn't affect performance because composition takes advantage of the pattern synonyms so it should be equal using `(+++)` or `(***)`

src/LinAlg.hs
conal: > Well `tr` is the beginning of the `tr`anspose word so I chose that abbreviation :stuck\_out\_tongue:

I’m guessing we all have the same idea of transposition as an operation on *matrices*, and maybe even come up with the implementation you gave for our compositional, generalized matrices `L`. I’m suggesting two fundamental questions:

  - What is the corresponding operation on *linear maps*?
  - Is transposition of conventional matrices or `L` consistent with the linear map operation? (In other words, is the semantic function a `tr`-homomorphism?)

src/LinAlg.hs
bolt12: Ups, you're right! What I really want is to use the Hadamar-Schur product instead of matrix composition! Now I understand your confusion! My bad!!

src/LinAlg.hs
bolt12: You can actually derive (it's a bit complicated) how the projections should look like from this `khatri fstM sndM == idL`, it's all describe in `Matrices as Arrows`.

bolt12:
> > Can I add the following type family (as described in “TYM”) to efficiently convert a TypeLit to a balanced tree?
> 
> I expect that this question will lead to an interesting and useful discussion of its own, including motivation and philosophy. I’d like to see it moved to a new issue, where it can be found and read more easily.

I'll move it to a new issue!

src/LinAlg.hs
conal: > @conal the standard definition of the Khatri-Rao product is presented as a (column-wise) variant of the Kronecker Product.

Okay. We could instead ask this denotation/specification question about the Kronecker product (as in issue \#6), which might be a simpler setting, and then tackle Khatri-Rao.

In my experience, most programmers and even applied mathematicians fail to clearly distinguish representations from the more abstract (and often simpler) mathematical objects being represented. Without this distinction, they often bake popular but arbitrary and unfortunate choices into their problem statement, eliminating the possibility of finding better questions with better answers. It can be frustrating to have familiar, commonly accepted answers close at hand and not just stick them in, but I have found it immensely rewarding to resist this temptation (and often pressure from colleagues or supervisors).

src/LinAlg.hs
bolt12: @conal I see. This is not an urgent matter so I do not think we should waste a lot of times with names, unless they are very important!

src/LinAlg.hs
conal: I don’t think a proof would be very difficult, but it would at least take a clear statement of the claim (what correctness means here) and then informal arguments about how the implementation achieves this claim.

src/LinAlg.hs
bolt12: I wish I could have a more detailed answer! But I'll keep trying to have better answers as I learn more!

src/LinAlg.hs
bolt12: @conal I'm trying to search the web and see if I can understand enough to give you an answer. Everywhere I look it says that [transposition is linear](https://www.quora.com/Is-transposition-a-linear-transformation) and we can actually think of this operation as applying a permutation matrix to other. Maybe I lack enough background about linear maps but from what I understand I don't think that the transposition of conventional matrices is consistent with the linear map operation (I might be wrong).

src/LinAlg.hs
conal: > The top matrix is the matrix full of ones, is a concept that comes from Relational Algebra, which is the largest Boolean matrix.

Oh, hm. Does such a matrix represent something interesting and/or useful in linear algebra?

For what ordering (on boolean matrices) is *top* maximal? Does this ordering generalize usefully from `Bool` to other semirings.

> `bang` must be a column vector so I do not think that it should be `U1`. I just gave it this name because my background comes from Algebra of Programming and a lot of the concepts transpose to linear algebra. Once we use quantitative matrices instead of boolean matrices the **top** matrix stops being the **top** matrix as known in Rel, obviously

Thanks. If the property motivating the name “top” doesn’t hold for linear algebra, wouldn’t we be better off renaming it to something true?

IIRC, the name “\!” comes from category theory where it refers to the unique morphism from any object to the terminal object. It’s a tricky name to search on, so I’m unsure. There’s a dual notion of initial arrow as well.

Since the type/object of scalars is *not* terminal in the category of linear maps, the name “\!”/“bang” is terribly misleading. It’s a great name to apply to linear maps to `U1`, however\!

bolt12: > For what ordering (on boolean matrices) is top maximal?

Relation inclusion!

> If the property motivating the name “top” doesn’t hold for linear algebra, wouldn’t we be better off renaming it to something true?

Yes absolutely, sorry giving it such a bad name!

src/LinAlg.hs
conal: I bet there's something interesting to be learned!

bolt12: I added Hadamard-Schur and fixed the khatri rao product. Now the type signature is not weird any more!

src/LinAlg.hs
bolt12: What should we do then?

src/LinAlg.hs
conal: Thanks for looking.

> Everywhere I look it says that [transposition is linear](https://www.quora.com/Is-transposition-a-linear-transformation) and we can actually think of this operation as applying a permutation matrix to other.

That’s exactly the sort of unsatisfying explanation I usually hear as well. Note that this “permutation” explanation is in terms of one particular *representation* of linear maps, not linear maps themselves. How do we know that there is a corresponding operation on the *meaning* of this representation? What is the correct corresponding operation on *different* representations?

Matrices are sometimes a very poor choice of representation, so these questions have practical importance\! For instance, decades of research on automatic differentiation (AD) and machine learning produced terribly complex implementations of efficient AD with complex and non-rigorous justifications. My recent simple implementation is more general and much simple in implementation, specification, and correctness proof, exactly because I chose a better linear map representation, which I could only do because I didn’t define the usual poor choice (matrices) into the problem statement. In doing this work, I finally came to understand transposition in terms of linear maps, and that understanding provided crucial insight leading to the simple, general algorithm and its simple proof.

The story is in [*The simple essence of automatic differentiation*](http://conal.net/papers/essence-of-ad), and I think it suggests a different name, description, and type for transposition.

src/LinAlg.hs
conal: Hm. A few questions:

  - Does anyone want to clearly say what it would mean for an `Eq (L f g s)` implementation to be correct?
  - Does anyone want to vouch for the correctness of this code according to an agreed-upon definition of correctness?
  - Are you (or is someone else) in a hurry to add this functionality? If so,
      - Would you like help with the proof (even if informal but beyond “it looks good”)?
      - If so, does someone want to volunteer to help resolve the correctness questions? I suspect that Dave and/or Daniel would enjoy the challenge. I can help, but I like seeing the rest of you flexing these muscles, and I bet you’d enjoy it and get experience valuable for similar, more difficult questions.

If you’re in a hurry to add the code, and no one volunteers to address correctness, you could add an issue for now, documenting the missing rigor and include the issue URL in a comment with the instance, so that (a) it doesn’t get forgotten and (b) we’ll have a clue where to look in case bugs surface.

Any other thoughts on how to proceed from here?

src/LinAlg.hs
```diff
@@ -194,20 +239,31 @@ instance (HasScaleV f, Semiring s) => Semiring (L f f s) where
...
+-- Hadamard-Schur product
+(.*) :: Semiring s => L f g s -> L f g s -> L f g s
+Zero .* Zero            = Zero
+Zero .* _               = Zero
+_ .* Zero               = Zero
+(Scale s) .* (Scale s') = Scale (s * s')
+(f :|# g) .* (h :| k)   = (f .* h) :| (g .* k)
+(f :&# g) .* (h :& k)   = (f .* h) :& (g .* k)
+ForkL ms .* Fork ms'    = ForkL $ liftR2 (.*) ms ms'
+JoinL ms .* Join ms'    = JoinL $ liftR2 (.*) ms ms'
```
conal, line 251: I’ve suspected for a while now that Hadamard-Schur products have very little to do with linear algebra. I’m much more skeptical about this one than about transposition or even Kronecker or Khatri-Rao products.

I think that first clause is redundant with the second and third.

I’d omit LHS parentheses in the `Scale` clause as you’ve done with the `ForkL` and `JoinL` clauses.

conal: My hunch is that whenever people use Hadamard-Schur products, their “matrices” are really 2D arrays representing something *other than* linear maps, and that it’s only lack of type discipline that allows their code (or writing) to type-check. I might be wrong and would be very interested to learn that I am.

src/LinAlg.hs
conal: BTW, the `Zero` constructor is adding subtlety (potential errors) and is not necessary for expressiveness. Maybe it should go\!

dwincort: My gut assumption would be that two linear maps `f` and `g` of type `a ⊸ b` are equal if, for all `x` of type `a`, `f x == g x`.  I'm not _exactly_ sure what this means for our `L` data type, but I think a value of type `L f g s` corresponds to a linear map `f s ⊸ g s`.  I think we can represent "linear map application" through multiplication (`.@`), so perhaps our domain and codomain types are more adequately described as something like `L Par1 f s` and `L Par1 g s` respectively.  (In matrix-talk, these would be vectors, so I think this makes sense.)

At this point, my argument feels a little circular because I'm saying that in order to define equality on `L f g s`, I need to test equality on `L Par1 g s`, so it would probably be nice to have an isomorphism between `L Par1 g s` and `g s` for arbitrary `g`, but that seems pretty straightforward to build.  Then, we could build a proof of equality structurally based on the definition of `.@`.

@conal, does that seem like the right track?

dwincort: > BTW, the `Zero` constructor is adding subtlety (potential errors) and is not necessary for expressiveness. Maybe it should go!

Sure!  We could easily make a `ConstV f g` type class with a function like `constV :: s -> L f g s` that makes a linear map where every element is the given `s` -- in fact, if we do want to keep Hadamar-Schur products, then the identity for such an operation would be easy to express as `constV one` (_not_ `scaleV one`).

Fewer constructors definitely makes pattern matching simpler.

conal: > @conal, does that seem like the right track?

Very much so\!

> I think we can represent “linear map application” through multiplication (`.@`) …
> 
> At this point, my argument feels a little circular because …

Indeed it does, or at least inelegant. I’d reconsider the choice to define application via composition.

conal: > … a function like `constV :: s -> L f g s` that makes a linear map where every element is the given `s`

What are *elements* of a linear map?

dwincort: > What are elements of a linear map?

I know, right?  _I don't know what that means._  It's purely an implementation-level detail (much like `diagR`) that we can use to construct values that represent the zero linear map.  Really, I should have said "that makes a **value of type `L f g s`** where every **component `Scale` constructor** is given `s`."  It's a super implementation-level specification and would potentially require a proof that `zero` really is the identity element of addition.

conal: All such software design questions become *much* clearer as soon as we have a precise denotation. With denotation in hand, it’s usually fairly obvious either what the answer is or that the question is malformed.

bolt12: Well I agree that less constructors is simpler. That was my approach in "TYM" too! But I think we can benefit from Zero in the long run (performance wise).

> Does anyone want to clearly say what it would mean for an `Eq (L f g s)` implementation to be correct?

I'm not very familiar with the linear map denotation but thinking in terms of usability I'd expect that the `Eq` instance on `L` would correctly be able to compare for equality two matrices regardless their permutation of Joins/Forks (hence using the pattern synonyms) and cannot distinguish `Scale 0` from `Zero`.

Regarding this, I think I can vouch for this implementation since it (declaratively) says the same as above.

> Are you (or is someone else) in a hurry to add this functionality?

I'm not in a hurry but this type class is useful to check if certain properties are true and check if the implementation of some operations are correct.

src/LinAlg.hs
bolt12: Yes you're right it is redundant, I'll fix it!

> I’ve suspected for a while now that Hadamard-Schur products have very little to do with linear algebra

The Hadamard-Schur product is basically the element-wise matrix multiplication. I noticed you have the same operation for the `Additive` type class:

```haskell
instance Additive s => Additive (L f g s) where
  zero = Zero
  Zero + m = m
  m + Zero = m
  Scale s + Scale s' = Scale (s + s') -- distributivity
  (f :|# g) + (h :| k) = (f + h) :| (g + k)
  (f :&# g) + (h :& k) = (f + h) :& (g + k)
  ForkL ms  + Fork ms' = Fork (ms +^ ms')
  JoinL ms  + Join ms' = Join (ms +^ ms')
```

How does addition relate to linear maps?

src/LinAlg.hs
conal: > It’s a super implementation-level specification …

Exactly. It’s an implementation detail that is specified only via other details of the same implementation.

bolt12: Hoping we can all agree on a definition for equality I propose the following:

- The `Zero` matrix needs to be undistinguishable from: `Scale 0`, `Zero :#& Zero`, `Zero :|& Zero`, `JoinL (Scale 0)`, `ForkL (Scale 0)`. 
- The `Scale s` matrix, given its type restrictions needs to be undistinguishable from: `Scale s`. But this raises some issues, because `Scale s` should be equal to `JoinL (Scale s)` or `ForkL (Scale s)` and I don't think the types allow this distinction (maybe I'm creating problems that don't exist), but `Par1 :.: Par1` is isomorphic to `Par1` and that can be an issue. The same can happen to the types `((Par1 :*: Par1) :*: Par1) ≃ (Par1 :*: (Par1 :*: Par1))` they represent matrices with the same dimensions but the type checker can not tell that. In my work on "TYM" I solved this issue by having a `Normalize` type family that when used with a new type wrapper it would abstract this issue away. Another way to solve this is to have a matrix that evidentiates the isomorphism between the types: an identity matrix of type `L ((Par1 :*: Par1) :*: Par1) (Par1 :*: (Par1 :*: Par1)) s`.
- If two matrices are equal but are defined in terms of different permutations of `(:#&)` and `(:|&)` or `JoinL` and `ForkL` these two matrices should not be distinguishable. Please note that given the different types we can run in the same issue as above: `(a :|& b)` should be equal to `JoinL (Par1 a :*: Par1 b)`.

Structural equality on our data type is quite complex mainly because the compiler is not capable of knowing that two types are isomorphic to each other. 

I'm not sure if the solution is elegant either. I happen to find the same issues in "TYM" and "fixed" them as I said.. which I agree is not the most elegant solution.

Is there any other alternative? Is there a notion of equality on linear maps?

conal: > I’m not in a hurry but this type class is useful to check if certain properties are true and check if the implementation of some operations are correct.

Good point about usefulness of an equality implementation for *testing* correctness.

This point also elevates the importance of deciding how we can assure that the equality implementation itself is correct\! Which raises the more fundamental questions of what `L` equality should mean and why.

conal: > Hoping we can all agree on a definition for equality I propose the following: \[…\]

Those three proposals are about a particular representation rather than about linear maps. I’d expect them all to be *consequences* of a simpler and more compelling definition of equality.

> Is there any other alternative? Is there a notion of equality on linear maps?

Yes\! Equality should always be denotational, i.e., two representations are considered equal exactly when their meanings are mathematically equal. This choice is a simple example of the semantic homomorphism (“precise analogy”) discipline used for *every* implemented operation, and failure is an abstraction leak as always. Equality is also an especially important case, because it is used pervasively in reasoning about all other operations. It also guarantees that algebraic/class laws transfer faithfully from the denotation/model to the representation/implementation.

bolt12: I remember from your talk about denotational design you had a `mu` function for the meaning of the operations. Here how should we setup such `mu` function? 

`mu ((==) m m') === ((==) (mu m) (mu m'))` seems obvious. And `mu m` should mean exactly our linear map. Should we have a linear map type? How can we achieve such equality definition?

conal: > I remember from your talk about denotational design you had a `mu` function for the meaning of the operations. Here how should we set up such `mu` function?

This is exactly the right question, to be addressed immediately after we have chosen a denotational model/type (linear maps) and a representation for that model (e.g., row-major matrices or our `L`). Everything else we do then follows.

> `mu ((==) m m') === ((==) (mu m) (mu m'))` seems obvious.

Yes, though aesthetically I prefer `mu (m == m') === (mu m == mu m')`.

> And `mu m` should mean exactly our linear map.

Yes. In other words, `mu` maps from `L` to its meaning.

> Should we have a linear map type?

Yes, spelled out precisely, whether executably or not. In particular, define equality extensionally as with general functions, via universal quantification over a typically uncountable set.

> How can we achieve such equality definition?

Given `mu` and the RHS `(==)`, the only unknown in this homomorphism equation is the LHS `(==)`. The computable solutions to this algebra problem are exactly the correct implementations of equality on `L`. Computable equality on a faithful representation of functions on uncountable types is one of the miracles of linear algebra and reflects the severity of linearity despite its tremendous usefulness.

src/LinAlg.hs
conal: Multiplication of functions does *not* preserve linearity, while addition and scalar multiplication (i.e., *linear* combination) does. Additional related comments at PR \#26.

src/LinAlg.hs
bolt12: I didn't quite understand how can we find what the LHS `(==)` should be, or you weren't suggesting it was findable in any way?

> Computable equality on a faithful representation of functions on uncountable types is one of the miracles of linear algebra and reflects the severity of linearity despite its tremendous usefulness.

Is this translatable to our data type?

conal: > I didn’t quite understand how can we find what the LHS `(==)` should be, or you weren’t suggesting it was findable in any way?

It is findable, but I wasn’t saying how. If you’d like to just update your first attempt, making the small changes I suggested above (removing `Scale` and some parenthesis), I think I’d be happy to merge. Or if you’d like, I’ll adapt your code and make the changes myself.

> > Computable equality on a faithful representation of functions on uncountable types is one of the miracles of linear algebra and reflects the severity of linearity despite its tremendous usefulness.
> 
> Is this translatable to our data type?

Yes.

bolt12: I will try and address the new version of `Eq` today!

bolt12:
I have made the changes! But I'm now in doubt if every function that needs to access the internal structure should be defined in terms of this isomorphism. Maybe it should, maybe it shouldn't because it might not be as performant...

Should `Eq` be instead defined as:

```haskell
instance (V f, ToRowMajor g, Eq (Rows f g s), Additive s) => Eq (L f g s) where
  a == b = lToRowMaj a == lToRowMaj b
```

Once again the `Additive` constraint, feels weird.

conal:
I like this one\! More succinctly via [`on`](https://hackage.haskell.org/package/base-4.14.0.0/docs/Data-Function.html#v:on),

``` haskell
  (==) = (==) `on` lToRowMaj
```

And no more need for `Additive s`.

bolt12:
@conal  I have made changes! Should I delete the Hadamard-Schur and Khatri-Rao product?

src/LinAlg.hs
```diff
@@ -173,20 +184,28 @@ instance (V f, Semiring s) => Semiring (L f f s) where
   one = idL
   (*) = (.@)
 
+-- Hadamard-Schur product
+(.*) :: Semiring s => L f g s -> L f g s -> L f g s
+Scale s .* Scale s'   = Scale (s * s')
+(f :|# g) .* (h :| k) = (f .* h) :| (g .* k)
+(f :&# g) .* (h :& k) = (f .* h) :& (g .* k)
+ForkL ms .* Fork ms'  = ForkL $ liftR2 (.*) ms ms'
+JoinL ms .* Join ms'  = JoinL $ liftR2 (.*) ms ms'
```
dwincort, line 193: I'm not saying we should keep the Hadamard-Schur product, but by the same logic that you used for `tr` (which, to be fair, @conal objects to), you could write this as:
```haskell
(.*) :: (V2 f g, Semiring s) => L f g s -> L f g s -> L f g s
a .* b = rowMajToL $ liftR2 (liftR2 (*)) (lToRowMaj a) (lToRowMaj b)
```

bolt12: Indeed! I understand that the isomorphism is only denotational and not representational/structural, and understand that this type of implementation takes advantage of the representational isomorphism and does not have any meaning denotationally and that might be a bad way to implement such operation. But why is `(==)` fine? Is there a way to "prove" that `(==)` via `rowMajtoL`/`lToRowMaj` is consistent with the linear map denotation? cc @conal

conal: Sorry I have not been responsive yesterday and today. I have been getting ready for my Haskell Love talk on Saturday.￼

conal: As I’ve mentioned elsewhere, `lToRowMaj` (being the inverse of `RowMajToL`) will flatten the `L` structure, so we’ll lose block computation in this composition and in whatever consumes the composition.

I also doubt we’ll want Hadamard-Schur products, but that question is still up in the air (see issues \#31, \#10, and \#6).

conal: H-SThe use of “`.`” in “`.*`” suggests a kind of composition to me, which I don’t think H-S is. In contrast, “`.@`” rightly suggests composition.

bolt12: @conal Don't worry about response time we are not in a hurry :smile:

> As I’ve mentioned elsewhere, lToRowMaj (being the inverse of RowMajToL) will flatten the L structure, so we’ll lose block computation in this composition and in whatever consumes the composition.

Not to mention the conversion overhead. So it seems to be more beneficial to use a specific implementation in terms of the matrix `L` structure, no? It also needs fewer constraints and that's always good

> But why is (==) fine? Is there a way to "prove" that (==) via rowMajtoL/lToRowMaj is consistent with the linear map denotation?

I still am confused about this.

> The use of “.” in “.*” suggests a kind of composition to me, which I don’t think suits H-S. (Do you?) In contrast, “.@” rightly suggests composition.

Yes I agree, but my imagination can not come up with a decent alternative, is there any suggestion? I'm happy to change it to something else!

dwincort: @conal I'll also chime in that there's no rush in responding (and apologies for the backlog of messages we're creating).  Good luck on your presentation tomorrow!

---

Quoting @conal from a couple of different places
> There are a lot of _transformations_ on linear maps, and many of those transformations are _isomorphisms_, but the join/fork exchange laws are _equalities_.

> I’m okay with adding `tr` for now, but I expect it will become a bug soon in which case we’ll need to remove or change it. Once the denotational model is in place, I want every operation on `L` to be defined by a corresponding operation on linear maps/functions. While transposition hints at a very important notion of duality in linear maps, I don’t think that notion naturally has the signature you’ve given.

> As I’ve mentioned elsewhere, `lToRowMaj` (being the inverse of `RowMajToL`) will flatten the `L` structure, so we’ll lose block computation in this composition and in whatever consumes the composition.

---

I think there are (at least) 3 different levels being talked about here: isomorphisms, denotational equality, and representational equality.  Isomorphisms and equality are connected by the fact that if `f` is an isomorphism between `X` and `Y` and `a,b : x` such that `a = b`, then `f a = f b`.  This follows directly from the fact that `f` defines a bijection between `X` and `Y`.  That said, denotational equality is a more generalized notion than representational equality, and `f` being a bijection under denotation does not imply that it's a bijection under representation.

Now, consider `RowMajToL` as our isomorphism, linear maps as our denotation, and `L` as our representation.  The join/fork exchange laws are _equalities_ under our denotation, but they're clearly not equalities of our representation.  Thus, `RowMajToL` functions as an isomorphism over our denotation but not our representation (the latter half of this is quite clear, as `RowMajToL` flattens `L`).  This is why we can use `RowMajToL` for our `Eq` instance.

My big question is kind of the inverse: it makes sense to me that we can use `RowMajToL` for `(==)`, but why are we concerned about using it for anything else.  Practically, I understand that it flattens the structure of `L`, but why does that matter if the result is correct, i.e., the result is exactly equal to a correct result?  It seems like we're trying to enforce that `tr` forms both sides of an isomorphism over _representational equality_ and not just denotational equality, but why do we care about this extra strength?  Why are we worrying about performance or representational equality in a function like `tr` when it hasn't been on our minds before?

Continuing this further, all of our definitions of binary operators have implicit "equality coercions" going on at the representational level.  For instance, consider two values `a,b : L f g s` that have different representations.  The result of `a + b` is not representationally equal to `b + a`, but we're fine with this.  Addition is commutative over the denotation but not the representation, and we don't care.

If I were to sum this all up in one sentence, I would say: I'm having trouble identifying where the "care"-"don't care" line is between denotation and representation.

bolt12: I think that's the core of my question too. Thanks for putting the effort in writing this!

> This is why we can use `RowMajToL` for our `Eq` instance.

So just to be sure. We can use `RowMajToL` because it is an isomorphism over our denotation so, in `mu (a == b) === mu a == mu b`, `lToRowMaj`/`rowToMajL` can be used because `mu (lToRowMaj a) == mu (lToRowMaj a)` ? If that's why then I also share your doubt about identifying where the "care"/"don't care" line is, because I think we can apply the same reasoning as `(==)` to `tr` and all other operations, right?

I'm sorry if I'm taking too long to understand all this.

conal: > > As I’ve mentioned elsewhere, `lToRowMaj` (being the inverse of `RowMajToL`) will flatten the `L` structure, so we’ll lose block computation in this composition and in whatever consumes the composition.
> 
> Not to mention the conversion overhead. So it seems to be more beneficial to use a specific implementation in terms of the matrix `L` structure, no?

The [context my remark](https://github.com/conal/linalg/pull/21#discussion_r463253826) was a suggestion about `(.@)` flattening its arguments. When @dwincort, @capn-freako, and I started talking about linear algebra a couple weeks ago, they were considering how to express block matrix multiplication. The join/fork algebra of linear (and other biproduct categories) gives a very nice answer. Coincidentally, I almost immediately met @bolt12, who’d been working on a very similar solution, leading us to this collaboration.

Given this history, I thought it worth pointing out that the flattening-based implementation of `(.@)` negates this initial motivation.

As a separate issue, for me the flattening reformulation lacks the beauty, insight, and novelty offered by the fork/join algebra.

> It also needs fewer constraints and that’s always good

I wouldn’t expect the constraints to differ. Do they?

conal: > > The use of “`.`” in “`.*`” suggests a kind of composition to me, which I don’t think suits H-S. (Do you?) In contrast, “`.@`” rightly suggests composition.
> 
> Yes I agree, but my imagination can not come up with a decent alternative, is there any suggestion? I’m happy to change it to something else\!

How about we move this question into issue \#31? If we discard the operation, we won’t need a name, and if we better understand its value (and essential nature), a good name may then suggest itself.

conal: > But why is `(==)` fine? Is there a way to “prove” that `(==)` via `rowMajtoL`/`lToRowMaj` is consistent with the linear map denotation?

Yes. A proof:

  - The corresponding property on denotations (linear maps/functions) holds.
  - `(==)` is (and must be) denotational.

We can discuss the why of each of these steps.

In denotational design, *every correctness proof* has this form.

conal:
> My big question is kind of the inverse: it makes sense to me that we can use `RowMajToL` for `(==)`, but why are we concerned about using it for anything else. Practically, I understand that it flattens the structure of `L`, but why does that matter if the result is correct, i.e., the result is exactly equal to a correct result? It seems like we’re trying to enforce that `tr` forms both sides of an isomorphism over *representational equality* and not just denotational equality, but why do we care about this extra strength? Why are we worrying about performance or representational equality in a function like `tr` when it hasn’t been on our minds before?

We can use `RowMajToL` for equality because doing do is correct (proof outline [above](https://github.com/conal/linalg/pull/21#discussion_r463914746)). We can use `RowMajToL` for other operations, such as composition for the same reason. My warnings against “flattening” in some cases was a warning to consider performance where I think we care about it, e.g., as in the efficient implementations @dwincort, @capn-freako, and I recently discussed. And not just about efficiency of one operation but about the operations consuming the result.

[My earlier remark about `tr`](https://github.com/conal/linalg/commit/630b99689d8b935506ae1fe6ab82e18077f8f35a#commitcomment-41011485) was something I thought @bolt12 might want to consider in evaluating his suggested redefinition of `tr`. If I have an objection to make, it’s again a loss of beauty and insight offered. The product/coproduct duality is reflected nicely in his earlier definition of `tr`, and this duality is a deep and wonderful thing. In other words, his earlier definition is a collection of *independently* true, useful, and lovely insights about not only linear functions/maps or even biproduct categories, but all cocartesian categories. In contrast, the terser version rolls all of these insights into a single wad, with the crucial monolithic operation `distribute` being on data structures.

I suggest keeping *both* implementations, because they both have their merits, and because their equivalence is also enlightening. We could do some performance measurements to choose the one called `tr` (but really given a better name and type signature), with the other given a similar name. Then add a theorem about their equivalence with a proof (on paper, or better, in Agda, Idris, Liquid Haskell, etc). Lacking a formal proof, we could at least add a QuickCheck property for systematic testing.

To understand what `tr` should really be called and what type it should have, consider what it means on linear functions rather than on matrices or on `L`.

src/LinAlg.hs
bolt12: > I wouldn’t expect the constraints to differ. Do they?

Well yes!

```haskell
instance (V f, ToRowMajor g, Eq (Rows f g s)) => Eq (L f g s) where
  a == b = lToRowMaj a == lToRowMaj b
```
needs `(V f, ToRowMajor g, Eq (Rows f g s))` and

```haskell
instance Eq s => Eq (L f g s) where
  Scale s == Scale s' = s == s'
  (f :|# g) == (h :| k)   = f == h && g == k
  (f :&# g) == (h :& k)   = f == h && g == k
  ForkL ms  == Fork ms'   = and $ liftR2 (==) ms ms'
  JoinL ms  == Join ms'   = and $ liftR2 (==) ms ms'
```

only needs `Eq s`

> (==) is (and must be) denotational.

Couldn't you say that, for example, the H-S product is (must be) denotational and then write it in terms of the `rowToMajL` denotational isomorphism?

bolt12:
@conal 

> My earlier remark about tr was something I thought @bolt12 might want to consider in evaluating his suggested redefinition of tr. If I have an objection to make, it’s again a loss of beauty and insight offered. The product/coproduct duality is reflected nicely in his earlier definition of tr, and this duality is a deep and wonderful thing. In other words, his earlier definition is a collection of independently true, useful, and lovely insights about not only linear functions/maps or even biproduct categories, but all cocartesian categories. In contrast, the terser version rolls all of these insights into a single wad, with the crucial monolithic operation distribute being on data structures.

> I suggest keeping both implementations, because they both have their merits, and because their equivalence is also enlightening. We could do some performance measurements to choose the one called tr (but really given a better name and type signature), with the other given a similar name. Then add a theorem about their equivalence with a proof (on paper, or better, in Agda, Idris, Liquid Haskell, etc). Lacking a formal proof, we could at least add a QuickCheck property for systematic testing.

I believe this also can apply to `(==)` so I pushed to this PR both versions of `tr` and `(==)`.

PR #26, opened 2020-07-28
No Zero constructor

conal:
This branch removes the `Zero :: L a b s` constructor for `L`. The tricky bit is synthesizing `zero :: L a b s` for all of the `a` and `b` we’ll want, which for now are built up from `Par1`, `(:*:)`, and `(:.:)`. This implementation adds one functor classes to make `L a Par1 s` using join and another for `L a b s` using fork.

Why eliminate `Zero`? To reduce cases considered by each operation, and especially reduce overlap.

We may want to reintroduce `Zero` later as an optimization. If so, we could add some optimizations we didn’t have previously, including binary fork or join of two zeros. I don’t know if n-ary join & fork of zeros is worth detecting.

### DIFF
--- a/src/LinAlg.hs
+++ b/src/LinAlg.hs
@@ -61,22 +61,6 @@ data L :: (* -> *) -> (* -> *) -> (* -> *) where
   JoinL :: V3 f g h => h (L f g s) -> L (h :.: f) g s
   ForkL :: V3 f g h => h (L f g s) -> L f (h :.: g) s
 
--- Scalable vectors
-class V a => HasScaleV a where
-  scaleV :: Semiring s => s -> L a a s
-
--- TODO: Can we eliminate HasScaleV in favor of using ToRowMajor?
-
-type HasScaleV2 a b = (HasScaleV a, HasScaleV b)
-
-instance HasScaleV Par1 where scaleV = Scale
-
-instance HasScaleV2 a b => HasScaleV (a :*: b) where
-  scaleV s = scaleV s *** scaleV s
-
-instance (HasScaleV2 a b, Representable b) => HasScaleV (b :.: a) where
-  scaleV s = cross (pureRep (scaleV s))
-
 unjoin2 :: (V3 f g h, Additive s) => L (f :*: g) h s -> L f h s :* L g h s
 unjoin2 (p :|# q) = (p,q)
 unjoin2 ((unjoin2 -> (p,q)) :&# (unjoin2 -> (r,s))) = (p :& r, q :& s)
@@ -172,7 +156,7 @@ diagRep :: (Representable h, Eq (Rep h)) => a -> h a -> h (h a)
 diagRep dflt as =
   tabulate (\ i -> tabulate (\ j -> if i == j then as `index` i else dflt))
 
-idL :: (HasScaleV a, Semiring s) => L a a s
+idL :: (V a, Semiring s) => L a a s
 idL = scaleV one
 
 infixr 9 .@
@@ -189,34 +173,34 @@ JoinL ms' .@ ForkL ms  = sum (ms' .^ ms)           -- biproduct law
      => p (L g h s) -> p (L f g s) -> p (L f h s)
 (.^) = liftR2 (.@)
 
-instance (HasScaleV f, Semiring s) => Semiring (L f f s) where
+instance (V f, Semiring s) => Semiring (L f f s) where
   one = idL
   (*) = (.@)
 
 -- Binary injections
 
-inl :: (HasScaleV a, V b, Semiring s) => L a (a :*: b) s 
+inl :: (V a, V b, Semiring s) => L a (a :*: b) s 
 inl = idL :& zero
 
-inr :: (V a, HasScaleV b, Semiring s) => L b (a :*: b) s 
+inr :: (V2 a b, Semiring s) => L b (a :*: b) s 
 inr = zero :& idL
 
 -- Binary projections
 
-exl :: (HasScaleV a, V b, Semiring s) => L (a :*: b) a s 
+exl :: (V2 a b, Semiring s) => L (a :*: b) a s 
 exl = idL :| zero
 
-exr :: (V a, HasScaleV b, Semiring s) => L (a :*: b) b s 
+exr :: (V2 a b, Semiring s) => L (a :*: b) b s 
 exr = zero :| idL
 
 -- Note that idL == inl :| inr == exl :& exr.
 
 -- N-ary injections
-ins :: (HasScaleV2 a c, Semiring s) => c (L a (c :.: a) s)
+ins :: (V2 a c, Semiring s) => c (L a (c :.: a) s)
 ins = unjoinL idL
 
 -- N-ary projections
-exs :: (HasScaleV2 a c, Semiring s) => c (L (c :.: a) a s)
+exs :: (V2 a c, Semiring s) => c (L (c :.: a) a s)
 exs = unforkL idL
 
 -- Note that idL == joinL ins == forkL exs
@@ -321,3 +305,6 @@ instance V2 a b => ToRowMajor (b :.: a) where
 zeroL :: (V2 a b, Additive s) => L a b s
 zeroL = rowMajToL (pureRep (pureRep zero))
 
+-- Vector scaling
+scaleV :: (V a, Semiring s) => s -> L a a s
+scaleV s = rowMajToL (diagRep zero (pureRep s))
--- a/src/LinAlg.hs
+++ b/src/LinAlg.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE UndecidableInstances #-} -- see below
+
 -- | Linear algebra after Fortran
 
 module LinAlg where
@@ -5,6 +7,7 @@ module LinAlg where
 import qualified Prelude as P
 import Prelude hiding ((+),sum,(*),unzip)
 
+import GHC.Types (Constraint)
 import GHC.Generics (Par1(..), (:*:)(..), (:.:)(..))
 import qualified Control.Arrow as A
 import Data.Distributive
@@ -19,9 +22,12 @@ type (:*)  = (,)
 unzip :: Functor f => f (a :* b) -> f a :* f b
 unzip ps = (fst <$> ps, snd <$> ps)
 
-type V f = (Representable f, Foldable f, Eq (Rep f))
+type V f = ((Representable f, Foldable f, Eq (Rep f), HasZA f, HasZB f) :: Constraint)
 type V2 f g = (V f, V g)
 type V3 f g h = (V2 f g, V h)
+type V4 f g h k = (V2 f g, V2 h k)
+
+-- Why does V suddenly need ":: Constraint" after adding HasZA f and HasZB f?
 
 class Additive a where
   infixl 6 +
@@ -52,10 +58,10 @@ infixr 2 :|#
 data L :: (* -> *) -> (* -> *) -> (* -> *) where
   -- Zero :: L f g s
   Scale :: s -> L Par1 Par1 s
-  (:|#) :: L f h s -> L g h s -> L (f :*: g) h s
-  (:&#) :: L f h s -> L f k s -> L f (h :*: k) s
-  JoinL :: V h => h (L f g s) -> L (h :.: f) g s
-  ForkL :: V h => h (L f g s) -> L f (h :.: g) s
+  (:|#) :: V3 f g h => L f h s -> L g h s -> L (f :*: g) h s
+  (:&#) :: V3 f h k => L f h s -> L f k s -> L f (h :*: k) s
+  JoinL :: V3 f g h => h (L f g s) -> L (h :.: f) g s
+  ForkL :: V3 f g h => h (L f g s) -> L f (h :.: g) s
 
 -- Scalable vectors
 class V a => HasScaleV a where
@@ -71,7 +77,7 @@ instance (HasScaleV a, HasScaleV b) => HasScaleV (a :*: b) where
 instance (HasScaleV a, HasScaleV b, Representable b) => HasScaleV (b :.: a) where
   scaleV s = cross (pureRep (scaleV s))
 
-unjoin2 :: Additive s => L (f :*: g) h s -> L f h s :* L g h s
+unjoin2 :: (V3 f g h, Additive s) => L (f :*: g) h s -> L f h s :* L g h s
 -- unjoin2 Zero = (zero,zero)
 unjoin2 (p :|# q) = (p,q)
 unjoin2 ((unjoin2 -> (p,q)) :&# (unjoin2 -> (r,s))) = (p :& r, q :& s)
@@ -91,7 +97,7 @@ unjoin2 (ForkL ms) = (ForkL A.*** ForkL) (unzip (unjoin2 <$> ms))
 (ForkL *** ForkL) (unzip (unjoin <$> ms)) :: L f (k :.: h) s :* L g (k :.: h) s
 #endif
 
-unfork2 :: Additive s => L f (h :*: k) s -> L f h s :* L f k s
+unfork2 :: (V3 f h k, Additive s) => L f (h :*: k) s -> L f h s :* L f k s
 -- unfork2 Zero = (zero,zero)
 unfork2 (p :&# q) = (p,q)
 unfork2 ((unfork2 -> (p,q)) :|# (unfork2 -> (r,s))) = (p :|# r, q :|# s)
@@ -99,11 +105,11 @@ unfork2 (JoinL ms) = (JoinL A.*** JoinL) (unzip (unfork2 <$> ms))
 
 -- unfork2 ((p :& q) :|# (r :& s)) = (p :|# r, q :|# s)
 
-pattern (:&) :: Additive s => L f h s -> L f k s -> L f (h :*: k) s
+pattern (:&) :: (V3 f h k, Additive s) => L f h s -> L f k s -> L f (h :*: k) s
 pattern u :& v <- (unfork2 -> (u,v)) where (:&) = (:&#)
 {-# complete (:&) #-}
 
-pattern (:|) :: Additive s => L f h s -> L g h s -> L (f :*: g) h s
+pattern (:|) :: (V3 f g h, Additive s) => L f h s -> L g h s -> L (f :*: g) h s
 pattern u :| v <- (unjoin2 -> (u,v)) where (:|) = (:|#)
 {-# complete (:|) #-}
 
@@ -111,7 +117,7 @@ pattern u :| v <- (unjoin2 -> (u,v)) where (:|) = (:|#)
 -- pattern Join ms <- (unjoinL -> ms) where Join = JoinL
 -- {-# complete Join #-}
 
-unforkL :: Representable h => L f (h :.: g) s -> h (L f g s)
+unforkL :: V3 f g h => L f (h :.: g) s -> h (L f g s)
 -- unforkL Zero       = pureRep Zero
 unforkL (p :|# q)  = liftR2 (:|#) (unforkL p) (unforkL q)
 unforkL (ForkL ms) = ms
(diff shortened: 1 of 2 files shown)


dwincort:
I wonder if adding generality would be helpful down the line.  For instance, instead of putting `zero` everywhere, you could take the `s` parameter as an argument.  In fact, if you took a function that generates `s` values, then it almost looks like you could make a `Representable` instance for `L f g` here.  Then, you could define `zeroL = tabulate $ const zero`, which would be pretty cool.

conal:
Thanks for the suggestion.

> `tabulate $ const zero`

or use [`pureRep`](https://hackage.haskell.org/package/adjunctions-4.4/docs/Data-Functor-Rep.html#v:pureRep) `zero`.

My `zeroL` definition denotes the zero linear map (the linear function that always yields zero). Would the generalized zero denote a linear map, and if so, which one? Similarly for a `Representable` instance.

I’m currently working on a precise denotational specification, which I hope we can all use for considering and discussing all API issues.

dwincort:
> My `zeroL` definition denotes the zero linear map (the linear function that always yields zero). Would the generalized zero denote a linear map, and if so, which one? Similarly for a `Representable` instance.

I'm a little confused by this question as I see this more as an implementation detail than a question of denotation.  I think we can agree that `zeroL` _should always_ denote the zero linear map however it is defined, but there are multiple ways to define it.  For instance, it could be a constructor `Zero :: L f g s` or a construction by using your `HasZA` and `HasZB` classes.  I'm merely proposing an alternate way to form that construction.

```haskell
class V a => RepA a where
  indexA :: L a Par1 s -> Rep a -> s
  tabulateA :: (Rep a -> s) -> L a Par1 s

class V b => RepB b where
  indexB :: RepA a => L a b s -> Rep b -> Rep a -> s
  tabulateB :: RepA a => (Rep b -> Rep a -> s) -> L a b s

instance RepA Par1 where
  indexA (Scale s) _ = s
  tabulateA f = Scale $ f ()

instance (RepA a, RepA a', V2 a a') => RepA (a :*: a') where
  indexA (a :| a') rep = either (indexA a) (indexA a') rep
  tabulateA f = tabulateA (f . Left) :| tabulateA (f . Right)

instance (RepA a, V2 c a) => RepA (c :.: a) where
  indexA (Join as) = uncurry $ indexA . index as
  tabulateA f = Join $ tabulate $ tabulateA . curry f

instance RepB Par1 where
  indexB = const . indexA
  tabulateB f = tabulateA $ f ()

instance (RepB b, RepB b', V2 b b') => RepB (b :*: b') where
  indexB (b :& b') = either (indexB b) (indexB b')
  tabulateB f = tabulateB (f . Left) :& tabulateB (f . Right)

instance (RepB b, V2 c b) => RepB (c :.: b) where
  indexB (Fork bs) = uncurry $ indexB . index bs
  tabulateB f = Fork $ tabulate $ tabulateB . curry f

instance V2 f g => Functor (L f g) where
  fmap f x = case x of
    Scale s -> Scale $ f s
    a :|# a' -> fmap f a :|# fmap f a'
    a :&# a' -> fmap f a :&# fmap f a'
    JoinL as -> JoinL $ fmap f <$> as
    ForkL as -> ForkL $ fmap f <$> as

instance (RepA f, RepB g, V2 f g) => Distributive (L f g) where
  distribute = distributeRep
  collect = collectRep

instance (RepA f, RepB g, V2 f g) => Representable (L f g) where
  type Rep (L f g) = (Rep g, Rep f)
  index = uncurry . indexB
  tabulate = tabulateB . curry
```

It seems to me that if `f` and `g` are `Representable`, it really should stand to reason that `L f g` is also `Representable`, and this is simply the definition of such.  As a perk, it means we can now define `zeroL = pureRep zero`.

(Interestingly, it's worth noting that this `Representable` instance wasn't possible while `Zero` was a constructor.)

conal:
Ah, interesting\!

Note that `(Rep a -> s) =~ a s`, which is the defining property of representable (naperian/trie) functors. Applying this isomorphism and renaming the classes and methods, we get two fundamental isomorphisms in linear algebra:

``` haskell
-- Duality isomorphism between row vectors and linear maps to scalars
class V a => HasDual a where
  vecToL :: a s -> L a Par1 s
  lToVec :: L a Par1 s -> a s

-- Row major isomorphism between matrices and linear maps
class V b => HasMatrix b where
  matToL :: HasDual a => b (a s) -> L a b s
  lToMat :: HasDual a => L a b s -> b (a s)
```

Note that `matToL` and `lToMat` are isomorphisms *denotationally*, note representationally, which is fine as long as we always specify and reason denotationally, including (especially) with equality.

Translating your suggestion, `zeroL = matToL (pureRep (pureRep zero))`.

There is an analogous pair of isomorphisms for column-major matrices.

conal:
We could also make the matrix functor more explicit:

``` haskell
class V b => HasMatrix b where
  matToL :: HasDual a => (b :.: a) s -> L a b s
  lToMat :: HasDual a => L a b s -> (b :.: a) s
```

or even

``` haskell
type Matrix a b = b :.: a  -- Row major

class V b => HasMatrix b where
  matToL :: HasDual a => Matrix a b s -> L a b s
  lToMat :: HasDual a => L a b s -> Matrix a b s
```

conal:
I agree that we could make a `Representable` instance on top of `HasMatrix` and the `Representable` instance for `Matrix` (noting that `Rep (b :.: a) = Rep b :* Rep a`).

My concern about this choice was that linear maps are functions, but the `Representable` on `L` would not match the representable for functions. Now I think a different lesson holds: *the representable for linear functions should not look like the representable for regular functions*, and instead should work like the representable for a matrix-like representation of linear functions.

I understand memoization as lossless conversion of functions into data structures (and back). If we want to efficiently memoize a *linear* function, we should take advantage linearity. If we don’t care about efficiency, we just use functions instead of representables (i.e., we wouldn’t memoize).

Memoization of other sets (categories?) of restricted functions should similarly consider how to memoize efficiently.

conal:
Some explanation of my creative process here, in case anyone is interested:

I did not brilliantly foresee that @dwincort’s generalization of my `zeroL` would become these two fundamental isomorphisms of linear algebra (nor the dual two isomorphisms for column vectors and column-major matrices). Rather, I got there by my habit of poking at familiar patterns and trying to refactor them away. It bothered me that `indexA` and `tabulateA` overlapped so much with `Representable`, so I wanted to factor out a use of `Representable` instead. A big clue was noticing `Rep a -> s`, which is isomorphic to `a s` via `Representable`.

Only after this refactoring did I see the two fundamental linear algebra isomorphisms. Then I saw the column-oriented isomorphisms, out of a general duality-seeking habit.

conal:
I’ll redo my `zeroL` implementation in this fashion, probably renaming the classes and methods to make the row-major aspect explicit and then add the column-major duals.

conal:
As I understand it, this duality between vectors and *linear functionals* (linear maps to scalars) is at the heart of what `transpose` means and what type to give it. (See [earlier discussion](https://github.com/conal/linalg/pull/21#pullrequestreview-455395357).) It’s also at the heart of simple, efficient, reverse-mode automatic differentiation ([*The simple essence of automatic differentiation*](http://conal.net/papers/essence-of-ad), sections 12 & 13).

conal:
Every element-wise operation on matrices must denote something (because of the isomorphism between linear maps and matrices), but it does *not* denote the same element-wise operation on the functions that the matrices denote, except in the very rare case of those operations being linear. The usual homomorphism properties thus fail. For this reason, we must take care in generalizing from addition to other binary operations, and from `const zero` to `const s` *any* other value of `s`. (The function `const s` is linear if and only if `s` is zero.)

If we do give a (sound) `Representable` instance to `L`, that instance will dictate instances for several other classes, including `Functor`, `Applicative`, `Monad`, and `Comonad`, all of which should be compatible with the corresponding instances for linear functions but not with regular functions. In particular, the `Functor` and `Applicative` instances will provide “element-wise” n-ary application of *arbitrary* (including nonlinear) functions. Should be very interesting\!

Other examples of inconsistency between linear functions and regular functions include everything to do with coproducts, since function coproducts are sums, while linear function coproducts are cartesian products (“direct sums”).

conal:
The latest commits add the row/linear-functional and linear-map/matrix isomorphisms described above and use it to define `zeroL` and `scaleV`. I like this code much better. Thanks for the prodding, @dwincort.

conal:
I'll merge this PR in a bit if I don't hear anything, but additional comments & questions are most welcome.

conal:
There are some more sweet uses of pattern synonyms here.

src/LinAlg.hs
```diff
@@ -91,28 +80,26 @@ unjoin2 (ForkL ms) = (ForkL A.*** ForkL) (unzip (unjoin2 <$> ms))
 (ForkL *** ForkL) (unzip (unjoin <$> ms)) :: L f (k :.: h) s :* L g (k :.: h) s
 #endif
 
-unfork2 :: Additive s => L f (h :*: k) s -> L f h s :* L f k s
-unfork2 Zero = (zero,zero)
+unfork2 :: (V3 f h k, Additive s) => L f (h :*: k) s -> L f h s :* L f k s
```
dwincort, line 83: Is `Additive s` still necessary?

src/LinAlg.hs
```diff
@@ -187,57 +169,144 @@ ForkL ms' .@ m         = ForkL (fmap (.@ m) ms')   -- n-ary product law
...
 
-instance (HasScaleV f, Semiring s) => Semiring (L f f s) where
+instance (V f, Semiring s) => Semiring (L f f s) where
   one = idL
   (*) = (.@)
 
 -- Binary injections
 
-inl :: (HasScaleV a, Semiring s) => L a (a :*: b) s 
+inl :: (V a, V b, Semiring s) => L a (a :*: b) s
```
dwincort, line 182: Just style, but maybe `V2 a b`?

conal:
I might be worth defining a category of isomorphisms with most of its vocabulary from category classes and then giving `ToScalar` and `ToRowMajor` each a single isomorphism-valued method instead of two mutually inverse methods. Isomorphisms form a monoidal category over products and sums, having `(+++)` and `(***)`. Isomorphisms cannot destroy or replicate information, however, so this category is neither cartesian nor cocartesian.

I expect isomorphisms to play an enormously important role in efficient implementation, including allowing us to program linear algebra in a lovely compositional style but still represent and manipulate efficiently via non-compositional arrays. This story is in [*Data-Parallel Programming without Arrays*](http://conal.net/papers/data-parallel-programming-without-arrays).

conal:
@bolt12: Now that the `Zero` constructor is gone and `Scale` is back to scalar only, I hope you’ll take another pass at adding an `Eq` instance, as in PR \#21. Correctness will be less subtle.

bolt12:
Wow! Super interesting developments! I'm excited with these changes! 

> If we do give a (sound) Representable instance to L, that instance will dictate instances for several other classes, including Functor, Applicative, Monad, and Comonad, all of which should be compatible with the corresponding instances for linear functions but not with regular functions. In particular, the Functor and Applicative instances will provide “element-wise” n-ary application of arbitrary (including nonlinear) functions. Should be very interesting!

Are you saying that we shouldn't give a `Representable` instance to `L` ? Or are you saying that if we do they will be consistent with the linear map denotation?

src/LinAlg.hs
```diff
@@ -187,57 +169,144 @@ ForkL ms' .@ m         = ForkL (fmap (.@ m) ms')   -- n-ary product law
...
+pattern LToRow m <- (rowToL -> m) where LToRow = lToRow
+{-# complete LToRow :: () #-}
+
+-- • A type signature must be provided for a set of polymorphic pattern synonyms.
+-- • In {-# complete LToRow #-}
+-- |
+-- | {-# complete LToRow #-}
+
+-- The ":: ()" gets around this error message, but I don't know whether the
+-- pattern will still work.
```
dwincort, line 250: I do believe that Class constraints can't be pattern matched like you're trying to do here.  I think your type signature is basically saying that the `LToRow` pattern is only applicable when `() ~ a s`.

More generally, why is a pattern necessary here?

bolt12:
> @bolt12: Now that the `Zero` constructor is gone and `Scale` is back to scalar only, I hope you’ll take another pass at adding an `Eq` instance, as in PR #21. Correctness will be less subtle.

@conal I will! Tomorrow I'll give it a go! I just had a busy day and am catching up on all these developments! I'm still a little confused about the `mu` function (as I said in #21 where I posed some questions regarding `JoinL`/`ForkL` and `Scale`)

src/LinAlg.hs
bolt12: I was going to say the same. This can be removed. I removed it in #21 if I'm not mistaken too.

conal:
> Are you saying that we shouldn’t give a `Representable` instance to `L` ? Or are you saying that if we do they will be consistent with the linear map denotation?

I’m saying the latter *and* that they should both be *inconsistent* with the `Representable` instance for regular functions. At first this discrepancy bothered me, but now I think is terrific, since it points to better forms of memoization in general, including matrices as a special case but far beyond.

src/LinAlg.hs
conal: Definitely. I like getting these details all neat and consistent.

src/LinAlg.hs
conal: > I do believe that Class constraints can’t be pattern matched like you’re trying to do here. I think your type signature is basically saying that the `LToRow` pattern is only applicable when `() ~ a s`.

I bet you’re right. Given the degree of polymorphism in the results of `LToRow` and to `lToRowMaj`, there isn’t a type constructor I can use here. Fortunately the inverses of these two patterns sufficed.

bolt12:
> My concern about this choice was that linear maps are functions, but the Representable on L would not match the representable for functions. Now I think a different lesson holds: the representable for linear functions should not look like the representable for regular functions, and instead should work like the representable for a matrix-like representation of linear functions.

Are you saying this because there's no other way of giving a sensible `Representable` instance `L`?

conal:
> I’m still a little confused about the `mu` function (as I said in \#21 where I posed some questions regarding `JoinL`/`ForkL` and `Scale`)

@bolt12: I’ve lost track of those specific questions, and I’d like to address them. Would you please link here to the specific comment(s) in \#21 (via the linked comment date, e.g. “2 days ago”)?

I think your original definitions are very close. You’ll need to drop clauses involving `Zero`, and I’d drop several unnecessary parentheses.

bolt12:
@conal [Here are the issues in question](https://github.com/conal/linalg/pull/21#discussion_r461095389)!

conal:
> Are you saying this because there’s no other way of giving a sensible `Representable` instance `L`?

Whatever operation we define on `L` must be explainable homomorphically in terms of the denotation of `L`, which I am choosing to be linear functions. This denotational homomorphic principle *is* the sensibility of my software design and implementation methodology. There maybe other “sensible” instance, but they would violate the basic sensibility on which I want this project to stand and thus lose the benefits.

We *could* decide that the `Representable` instance *for both* `L` and its denotation (linear functions) work like regular functions, but

  - Most uses of linear algebra involve *continuous domains* like the real- or complex-valued scalars and n-tuples (or more generally representable functors) of those scalars. Memoizing over functions of continuous domains is a research problem. It’s rarely pointed out, but functional memoization is correct only for “hyperstrict” functions, i.e., functions that map any “incomplete” values (containing bottom anywhere) and therefore any infinitely represented values to bottom. See [*Denotational design with type class morphisms*](http://conal.net/papers/type-class-morphisms/), section 9. Continuous types like the reals have value representations with infinite detail, so memoization would replace almost all interesting computations with bottom (operationally, non-termination). I think I do have a very nice, solution rooted in elegant domain theory (the math underpinning denotational semantics), but it goes beyond Haskell and other functional languages ([*Non-strict memoization*](http://conal.net/blog/posts/nonstrict-memoization)). Even if we could memoize functions on continuous/infinite values, we should in theory almost never get a cache hit. If arguments are uniformly distributed (unlikely), the probability of *ever* getting the same argument twice is zero (though not impossible).
  - Instead of tacking this difficult and interesting research memoization problem and then getting little or no benefit, we can instead use a different means of memoization that avoids the difficulties of *continuity* and exploits the restriction to *linearity*, leading to a compact and efficient representation with no loss of information. The result is something like *matrices* in either its conventional, non-compositional, non-type-safe form, or in our compositional and type-safe form (`L`).

conal:
> @conal [Here are the issues in question](https://github.com/conal/linalg/pull/21#discussion_r461095389)\!

That link takes me to your comment beginning as follows:

> Hoping we can all agree on a definition for equality I propose the following:

I think I responded, saying that equality on `L` must denote equality on meanings, i.e., on functions. Equality on linear functions/maps will be exactly as on regular functions: equal when they are equal for all arguments (usually uncountably many in linear algebra). Because of linearity, however, one can prove (not define) that linear functions are equal exactly when their matrix representations are equal. We’re using a more flexible representation (as you did for LAOP & *Type Your Matrices*), so we must take the fork & join equations into account, which your `(==)` definition did.

Maybe it will become clearer when I write down the semantic function (mapping `L` to linear functions).

Now I see [a newer comment from you](https://github.com/conal/linalg/pull/21#discussion_r461948128), to which I’ll respond.

conal:
Thanks for the very helpful discussions!

bolt12:
@conal thank you for all your helpful answers!

Regarding my issue maybe I should have just quoted it here:

> The `Scale s` matrix, given its type restrictions needs to be undistinguishable from: `Scale s`. But this raises some issues, because `Scale s` should be equal to `JoinL (Scale s)` or `ForkL (Scale s)` and I don't think the types allow this distinction (maybe I'm creating problems that don't exist), but `Par1 :.: Par1` is isomorphic to `Par1` and that can be an issue. The same can happen to the types `((Par1 :*: Par1) :*: Par1) ≃ (Par1 :*: (Par1 :*: Par1))` they represent matrices with the same dimensions but the type checker can not tell that. In my work on "TYM" I solved this issue by having a `Normalize` type family that when used with a new type wrapper it would abstract this issue away. Another way to solve this is to have a matrix that evidentiates the isomorphism between the types: an identity matrix of type `L ((Par1 :*: Par1) :*: Par1) (Par1 :*: (Par1 :*: Par1)) s`.

My question is then about this. Should our `Eq` instance distinguish between `Scale s` and `JoinL (Scale s)`. Given that the types are different, `Eq` says they are not comparable. But the types are isomorphic and should be comparable in some sense (if we want to, basically). I'm "TYM" I solved this issue by having new type wrappers and type families. Is this something we'll want to address, does this also make part of the API design and denotational design? I'd say no, but I want to know your opinion!

bolt12:
@conal  Why is `Additive` needed here:

```haskell
class ToScalar a where
  rowToL :: Additive s => a s -> L a Par1 s
  lToRow :: Additive s => L a Par1 s -> a s

class ToRowMajor b where
  rowMajToL :: (V a, Additive s) => Rows a b s -> L a b s
  lToRowMaj :: (V a, Additive s) => L a b s -> Rows a b s
```

?

Also I'm confused because you say that these two type classes describe the row/linear-functional and linear-map/matrix isomorphisms and you're able to `zeroL` which is the zero linear map. And what I understood from another thing you said 

> For this reason, we must take care in generalizing from addition to other binary operations, and from const zero to const s any other value of s. (The function const s is linear if and only if s is zero.)

was that only this `zeroL` is correct wrt to our denotation, but now we can implement `anyL` which might not be linear (?) as I did in #21 with `ones`.

conal:
I just removed `Additive` in `ToScalar` or `ToRowMajor` and other places in PR \#30. Thanks. With this change `Additive` won’t be needed by `tr` either.

> … was that only this `zeroL` is correct wrt to our denotation, but now we can implement `anyL` which might not be linear (?) as I did in \#21 with `ones`.

I’ll assume that by `anyL s` you mean a “matrix” with every cell equal to `s`. That matrix does denote a linear map (since every matrix does), but it cannot possibly denote `const s` unless `s = zero`, because no matrix can denote a nonlinear function, and `const s` is linear if and only if `s = zero`. Likewise, element-wise (Hadamard-Schur) matrix multiplication produces some matrix, which thus must denote a linear map, but it cannot denote the product of the linear functions denoted by the matrix factors, because linearity is not closed under multiplication. I don’t know whether there are any natural uses of Hadamard-Schur multiplication, i.e., uses inference system matrices are being used to denote linear maps rather than some other interpretation of 2D arrays.

bolt12:
I see, it makes sense to me now! Thanks for explaining!

Issue #23, opened 2020-07-27
Failable Pattern Synonyms


dwincort:
I've been playing around with the current work, and I find myself a bit frustrated with the pattern synonyms.  They're simply not as strong as the actual constructors because they have type requirements.  Consider that a value of type `L f g s` can always be matched against `JoinL ms`, but the same cannot be said for the pattern `Join ms` — trying yields an error like:
```
Couldn't match type ‘g’ with ‘h0 :.: f0’
```
It finally dawned on me that this is connected to the fact that every pattern synonym is individually marked as complete, something that struck me as strange at first.

Really, what I want is a set of patterns that _altogether_ are complete over the type `L f g s`, but individually may each fail to match.  Unfortunately, this is easier said than done.  I took a stab at it but only got partway there before getting stuck.  I'm using this issue to track the work done so far and see if anyone has any suggestions.  Regardless, I may keep working on it.

---

I arbitrarily started with `Join`.  I want a pattern synonym that will match `L f g s` exactly in the case where `f` is of the form `(h :.: f')` for some `h` and `f'` _as dictated by the type of `f`_ and not match otherwise.  My approach is a bit convoluted, but it's the only thing I got (even partially) to work.

We start by defining a GADT that carries information about whether there is a join or not:
```haskell
data IsJoin f g s where
  NotJoin :: IsJoin f g s
  IsJoin :: V h => h (L f g s) -> IsJoin (h :.: f) g s
```
along with a function:
```haskell
isJoin :: L f g s -> IsJoin f g s
```
The idea here is that if `isJoin` returns a value that is a `IsJoin`, then we correctly have a `Join`, and if not, we don't.  Before writing the definition of `isJoin`, we can indeed already write our pattern:
```haskell
pattern Join' :: () => (f ~ (h :.: f'), V h) => h (L f' g s) -> L f g s
pattern Join' ms <- (isJoin -> IsJoin ms) where Join' = JoinL
```
[...]

PR #27, opened 2020-07-28
Add Failable Patterns

dwincort:
These don't work exactly how I expected them to, but I'm opening the PR anyway for discussion.

## Overview

This PR adds four "failable" patterns for the four interesting constructors of `L`.  In order to not have any naming conflicts, I named them  `(:|:)`, `(:&:)`, `Join'`, and `Fork'`.

A failable pattern is one that _may_ not match.  For instance, a value of type `L f g s` _might_ indeed be an n-ary join, but since we don't know what `f` is, we don't know for sure.  Using the `Join'` pattern is much like using the `JoinL` pattern in that it may not match, but if it does, then we are guaranteed that there exists some `h` and `f'` such that `f ~ (h :.: f')`.  Because of this, we can use the `Fork'` pattern in more places than we can use the `Fork` pattern and not get annoying type errors.

One repercussion of failable patterns is that no individual pattern is "complete".  Rather, we can only guarantee a match if we try all 4 new patterns together (along with `Scale` and `Zero`).  This is actually quite sensible and not a particular downside.

In short, these failable patterns are the ideal for what we want when pattern-matching a value of type `L f g s`.  They're more general than the constructors (`Fork'` may match even if `ForkL` doesn't) but not type-restricted like the current patterns.

## Downsides

There are 4 major downsides of these failable patterns.  Any one of them might be a showstopper, and I'm not sure they're all fixable.

### Zero

Because the failable patterns are not type-restricted like the current pattern synonyms are, they must instead be type-_directed_.  That is, matching or failing to match must be determined by the types, and the constraint that the pattern produces (like `f ~ (h :.: f')` in my above example for `Join'`) must be directed from the types.  This is fine for a normal join or fork — using the GADT constructors _is_ type-directed, and matching on, say, `JoinL` yields the types `h` and `f'` that we need — but it is impossible for `Zero`.  In short, I'm not sure it's possible to support these failable patterns while `Zero` is a constructor and still have them do the "right thing".

There are two solutions.  One is remove the `Zero` constructor.  This may be bad for performance/optimizations, but it eases reasoning now.  The other is to declare that `Zero` cannot be matched by any of these new failable patterns.  In practice, this is not a problem (just match against `Zero` first any time you're doing pattern matching), but the theory is not so nice.

### Ugly Implementation

This is the implementation I came up with for the pattern `Join'` when the constructor is a `ForkL`
```haskell
isJoin :: forall f g s. V2 f g => L f g s -> Maybe (IsJoin f g s)
isJoin (ForkL (ms :: k (L f g' s))) = case sequenceA (isJoin <$> ms) of
  Just ys -> case ifoldrRep go (unsafeCoerce $ Dict @(), Nothing, const undefined) ys of
    (Dict, Just Dict, k) -> Just $ IsJoin $ fmap ForkL $ distribute $ tabulate k
    _ -> Nothing
  Nothing -> Nothing
  where
    go :: Rep k -> IsJoin f g' s -> (Dict (f ~ (h :.: f')), Maybe (Dict (V2 f' h)), Rep k -> h (L f' g' s))
       -> (Dict (f ~ (h :.: f')), Maybe (Dict (V2 f' h)), Rep k -> h (L f' g' s))
    go k (IsJoin hs) (Dict, _, fun) = (Dict, Just Dict, \k' -> if k == k' then hs else fun k')
```
Ooof!  Not only did I have to add `Traversable` to `V`, implement a new `ifoldrRep` function, and bring in `Dict` from the `constraints` package, but I _still_ end up using `unsafeCoerce` and `undefined`.  All that said, I'm quite confident it does the right thing.  For instance, the `undefined` should be impossible to access, and the `unsafeCoerce` ends up being guarded.  (Also, I think the `undefined` could be replaced by a clever use of `fix`.)

### Bad Type-Checker Performance

I have no idea why, but when I add the line
```haskell
{-# COMPLETE Zero, Scale, (:&:), (:|:), Join', Fork' #-}
```
to the module, type checking slows to a crawl.  In fact, I thought I may have put it into an infinite loop, but after waiting long enough, it really did finish.  I wonder if others can reproduce this.

### Insufficient Complete Pragmas

The new patterns are a little frustrating because while they're ideal for matching against an arbitrary value of type `L f g s`, they suffer from the same drawbacks as the regular constructors when matching against values with more constrained types.  For instance, we know that a value of type `L (h :.: f) g s` will always successfully match to the pattern `Join' ms`, but as far as I can tell, there's no way to tell GHC this fact.  This is pretty frustrating as it means that a double pattern match will yield "non-exhaustive pattern match" errors even when we know the match is exhaustive.  For instance, consider a case in the definition of `(+)`:
```haskell
  Join' ms  + Join' ms' = Join' (ms +^ ms')
```
If the first argument matches, we know that the second argument will also match, but GHC does not.  It will instead argue that we should provide two more cases, of the forms
```haskell
  (Join' _) + (Fork' _)
  (Join' _) + (_ :&: _)
```
"just in case" the second argument turned out not to match `Join'`.

The easy way to fix this is to use type-restricted patterns on the right-hand side, but this means we need to keep around all of our type-restricted patterns.  The question then becomes, what have these failable patterns actually given us?

## Failable Patterns in Practice

The main benefit of failable patterns is that if a value is inherently, say, a fork, then it will always succeed when matching against `Fork'`.  This is a great power, but it does seem a little weird too.  Consider the following operation:
```haskell
foo :: L f g s -> String
foo (ForkL _) = "Fork"
foo (JoinL _) = "Join"
foo _         = "Other"
```
Also consider a value `v :: L (c :.: a) (c :.: b) s` for some `a`, `b`, and `c`.  What should `foo v` be?  Well, it will depend on exactly how `foo` was constructed.  This means that `foo` is _data-directed_.  Now consider these two alternatives:
```haskell
fooF :: L f g s -> String
fooF (Fork' _) = "Fork"
fooF (Join' _) = "Join"
fooF _         = "Other"

fooJ :: L f g s -> String
fooJ (Join' _) = "Join"
fooJ (Fork' _) = "Fork"
fooJ _         = "Other"
```
Regardless of how we constructed `v`, `fooF v` will always be `"Fork"` and `fooJ v` will always be `"Join"`.  Our functions are no longer data driven, and indeed, the order of the lines of the definition has a very real impact on the result of the function.  Even for functions that preserve semantic behavior in all match orderings, there may be performance effects.

I'm not saying this is inherently a bad or good thing, but it is definitely notable.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 linalg.cabal
+.stack*
--- a/package.yaml
+++ b/package.yaml
@@ -29,7 +29,7 @@ default-extensions:
   - TypeOperators
   # - UndecidableInstances
   - ViewPatterns
-  # - ScopedTypeVariables
+  - ScopedTypeVariables
   - KindSignatures
   # - NoStarIsType
   - TypeSynonymInstances
@@ -42,6 +42,7 @@ library:
     - base
     - distributive
     - adjunctions
+    - constraints
   other-modules: []
   source-dirs: src
   exposed-modules:
--- a/src/LinAlg.hs
+++ b/src/LinAlg.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE UndecidableInstances #-}
 -- | Linear algebra after Fortran
 
 module LinAlg where
@@ -9,6 +10,8 @@ import GHC.Generics (Par1(..), (:*:)(..), (:.:)(..))
 import qualified Control.Arrow as A
 import Data.Distributive
 import Data.Functor.Rep
+import Data.Constraint (Dict(..))
+import Unsafe.Coerce (unsafeCoerce)
 
 infixl 7 :*
 type (:*)  = (,)
@@ -19,7 +22,7 @@ type (:*)  = (,)
 unzip :: Functor f => f (a :* b) -> f a :* f b
 unzip ps = (fst <$> ps, snd <$> ps)
 
-type V f = (Representable f, Foldable f, Eq (Rep f))
+type V f = (Representable f, Foldable f, Eq (Rep f), Traversable f)
 type V2 f g = (V f, V g)
 type V3 f g h = (V2 f g, V h)
 
@@ -50,12 +53,12 @@ infixr 2 :|#
 -- | Compositional Linear map representation. @L f g s@ denotes @f s -* g s@,
 -- where @(-*)@ means linear functions.
 data L :: (* -> *) -> (* -> *) -> (* -> *) where
-  Zero :: L f g s
+  Zero :: V2 f g => L f g s
   Scale :: s -> L Par1 Par1 s
-  (:|#) :: L f h s -> L g h s -> L (f :*: g) h s
-  (:&#) :: L f h s -> L f k s -> L f (h :*: k) s
-  JoinL :: V h => h (L f g s) -> L (h :.: f) g s
-  ForkL :: V h => h (L f g s) -> L f (h :.: g) s
+  (:|#) :: V3 f g h => L f h s -> L g h s -> L (f :*: g) h s
+  (:&#) :: V3 f g h => L f g s -> L f h s -> L f (g :*: h) s
+  JoinL :: V3 f g h => h (L f g s) -> L (h :.: f) g s
+  ForkL :: V3 f g h => h (L f g s) -> L f (h :.: g) s
 
 -- Scalable vectors
 class V a => HasScaleV a where
@@ -71,7 +74,7 @@ instance (HasScaleV a, HasScaleV b) => HasScaleV (a :*: b) where
 instance (HasScaleV a, HasScaleV b, Representable b) => HasScaleV (b :.: a) where
   scaleV s = cross (pureRep (scaleV s))
 
-unjoin2 :: Additive s => L (f :*: g) h s -> L f h s :* L g h s
+unjoin2 :: (Additive s, V3 f g h) => L (f :*: g) h s -> L f h s :* L g h s
 unjoin2 Zero = (zero,zero)
 unjoin2 (p :|# q) = (p,q)
 unjoin2 ((unjoin2 -> (p,q)) :&# (unjoin2 -> (r,s))) = (p :& r, q :& s)
@@ -91,7 +94,7 @@ unjoin2 (ForkL ms) = (ForkL A.*** ForkL) (unzip (unjoin2 <$> ms))
 (ForkL *** ForkL) (unzip (unjoin <$> ms)) :: L f (k :.: h) s :* L g (k :.: h) s
 #endif
 
-unfork2 :: Additive s => L f (h :*: k) s -> L f h s :* L f k s
+unfork2 :: (V3 f g h, Additive s) => L f (g :*: h) s -> L f g s :* L f h s
 unfork2 Zero = (zero,zero)
 unfork2 (p :&# q) = (p,q)
 unfork2 ((unfork2 -> (p,q)) :|# (unfork2 -> (r,s))) = (p :|# r, q :|# s)
@@ -99,11 +102,11 @@ unfork2 (JoinL ms) = (JoinL A.*** JoinL) (unzip (unfork2 <$> ms))
 
 -- unfork2 ((p :& q) :|# (r :& s)) = (p :|# r, q :|# s)
 
-pattern (:&) :: Additive s => L f h s -> L f k s -> L f (h :*: k) s
+pattern (:&) :: (V3 f g h, Additive s) => L f g s -> L f h s -> L f (g :*: h) s
 pattern u :& v <- (unfork2 -> (u,v)) where (:&) = (:&#)
 {-# complete (:&) #-}
 
-pattern (:|) :: Additive s => L f h s -> L g h s -> L (f :*: g) h s
+pattern (:|) :: (V3 f g h, Additive s) => L f h s -> L g h s -> L (f :*: g) h s
 pattern u :| v <- (unjoin2 -> (u,v)) where (:|) = (:|#)
 {-# complete (:|) #-}
 
@@ -111,7 +114,7 @@ pattern u :| v <- (unjoin2 -> (u,v)) where (:|) = (:|#)
 -- pattern Join ms <- (unjoinL -> ms) where Join = JoinL
 -- {-# complete Join #-}
 
-unforkL :: Representable h => L f (h :.: g) s -> h (L f g s)
+unforkL :: (V3 f g h) => L f (h :.: g) s -> h (L f g s)
 unforkL Zero       = pureRep Zero
 unforkL (p :|# q)  = liftR2 (:|#) (unforkL p) (unforkL q)
 unforkL (ForkL ms) = ms
@@ -136,21 +139,21 @@ liftR2 (:|#) (unforkL p) (unforkL p') :: h (L (f :*: f') g s)
 JoinL <$> distrib (unforkL <$> ms) :: h (L (k :.: f) g s)
 #endif
 
-unjoinL :: Representable h => L (h :.: f) g s -> h (L f g s)
+unjoinL :: (V3 f g h) => L (h :.: f) g s -> h (L f g s)
 unjoinL Zero       = pureRep Zero
 unjoinL (p :&# p') = liftR2 (:&#) (unjoinL p) (unjoinL p')
 unjoinL (JoinL ms) = ms
 unjoinL (ForkL ms) = fmap ForkL (distribute (fmap unjoinL ms))
 
-pattern Fork :: V h => h (L f g s) -> L f (h :.: g) s
+pattern Fork :: V3 f g h => h (L f g s) -> L f (h :.: g) s
 pattern Fork ms <- (unforkL -> ms) where Fork = ForkL
 {-# complete Fork #-}
 
-pattern Join :: V h => h (L f g s) -> L (h :.: f) g s
+pattern Join :: V3 f g h => h (L f g s) -> L (h :.: f) g s
 pattern Join ms <- (unjoinL -> ms) where Join = JoinL
 {-# complete Join #-}
 
-instance Additive s => Additive (L f g s) where
+instance (V2 f g, Additive s) => Additive (L f g s) where
   zero = Zero
   Zero + m = m
   m + Zero = m
@@ -176,7 +179,7 @@ idL :: (HasScaleV a, Semiring s) => L a a s
 idL = scaleV one
 
 infixr 9 .@
-(.@) :: Semiring s => L g h s -> L f g s -> L f h s
+(.@) :: (V3 f g h, Semiring s) => L g h s -> L f g s -> L f h s
 Zero      .@ _         = Zero                      -- Zero denotation
 _         .@ Zero      = Zero                      -- linearity
 Scale a   .@ Scale b   = Scale (a * b)             -- Scale denotation
@@ -187,7 +190,7 @@ ForkL ms' .@ m         = ForkL (fmap (.@ m) ms')   -- n-ary product law
 m'        .@ JoinL ms  = JoinL (fmap (m' .@) ms)   -- n-ary coproduct law
 JoinL ms' .@ ForkL ms  = sum (ms' .^ ms)           -- biproduct law
 
-(.^) :: (Representable p, Semiring s) => p (L g h s) -> p (L f g s) -> p (L f h s)
+(.^) :: (Representable p, V3 f g h, Semiring s) => p (L g h s) -> p (L f g s) -> p (L f h s)
 (.^) = liftR2 (.@)
 
 instance (HasScaleV f, Semiring s) => Semiring (L f f s) where
@@ -196,18 +199,18 @@ instance (HasScaleV f, Semiring s) => Semiring (L f f s) where
 
 -- Binary injections
 
-inl :: (HasScaleV a, Semiring s) => L a (a :*: b) s 
+inl :: (HasScaleV a, V b, Semiring s) => L a (a :*: b) s
 inl = idL :& zero
 
-inr :: (HasScaleV b, Semiring s) => L b (a :*: b) s 
+inr :: (HasScaleV b, V a, Semiring s) => L b (a :*: b) s
 inr = zero :& idL
 
 -- Binary projections
 
-exl :: (HasScaleV a, Semiring s) => L (a :*: b) a s 
+exl :: (HasScaleV a, V b, Semiring s) => L (a :*: b) a s
 exl = idL :| zero
 
-exr :: (HasScaleV b, Semiring s) => L (a :*: b) b s 
+exr :: (HasScaleV b, V a, Semiring s) => L (a :*: b) b s
 exr = zero :| idL
 
 -- Note that idL == inl :| inr == exl :& exr.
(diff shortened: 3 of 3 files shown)


conal:
For cross, referencing, see issue #23.

conal:
Wow. That looks like a hell of a battle. Congratulations.

Here are some questions, in addition to the ones you raised:

  - Could the essential lessons be factored out so that they could be read, understood, and reasoned about independently of what we’re doing here?
  - Could the code be tamed into simplicity of precise human and automated reasoning? (I intend to target our implementation at non-CPU architectures, which means bypassing GHC at the back end.)
  - How important do you now see the problem being solved here?
  - What’s your current sense of where you’d like to go with what you’ve learned here, in addition to a learning experience (for me as well)?

dwincort:
> - Could the essential lessons be factored out so that they could be read, understood, and reasoned about independently of what we’re doing here?
> - Could the code be tamed into simplicity for practical, precise human and automated reasoning? (I intend to target our implementation at non-CPU architectures, which means bypassing GHC at the back end.)
> - How important do you now see the problem being solved here?
> - What’s your current sense of where you’d like to go with what you’ve learned here, in addition to a learning experience for all of us?

I think nearly all of these questions can be answered together by splitting the work I did for this PR into two parts: the technical and the theoretical.

### The Technical

A lot of the hard work was convincing the type checker to know what I know, and I accomplished this in a lot of hacky ways.  Each of those hacks can be teased apart and discussed individually, and now that I think about it, it might be fun to do a write-up of this.  I wouldn't call these hacks good practice ordinarily, but they do have their place.

Furthermore, part of the reason it's so messy is that I tried to bury the hacks so that the type checker really was doing the majority of the work.  In other words, I was trying to highlight (as best I could) the proof part and have the `unsafeCoerce` and `undefined` be relegated to being almost inconsequential.  In practice, I could lift some of the burden from the type checker and use `unsafeCoerce` more liberally without changing the behavior of the function.

One thing I discovered from this effort is that GHC's pattern synonym solution — particularly the complete pragma — isn't quite as strong as I want it to be.  The bad type-checker performance seems like a bug that I'd like to try to reproduce more simply, and the issue I highlighted in "Insufficient Complete Pragmas" may not be a bug per se, but deserves to be made into a GHC issue if it isn't one already.  (In fact, I should probably check if the bad performance has already been reported before I try to report it myself.)

### The Theoretical

I've been wrestling a lot with the topic I brought up in "Failable Patterns in Practice", and I think it's related to @bolt12's questions about `FromNat` and isomorphisms (e.g., the difference between `Par1 (Par1 s)` and `Par1 s`).  Honestly, it makes me wonder if I didn't go far enough with the failable patterns.

We already accept that if we have a fork of joins then that should be equivalent to a join of forks.  Indeed, the pattern synonyms @conal already has do this.  But, what if we have a join of joins — can we reorder those joins?  I feel like we should be able to (and I don't think it matters if the joins are both binary, both n-ary, or one of each).

Can we do even more?  What if we have a binary join of two things that happen to be the same — can we change that into an n-ary join where our representable functor has cardinality two?  This question is the big one and really gets to the heart of my wrestling.  If we think of the things we're manipulating as matrices, then easily, the answer should be yes.  But should it be true for linear maps in general?  I can't come up with a counterexample, but that may be a failure of imagination on my part.

Why does this matter?  It matters precisely because of the difference in behavior I describe above between `fooF` and `fooJ`.  Why does it make sense at all to pattern match on a constructor like `ForkL` or `JoinL` when the two representations may be isomorphic?  Isn't this tantamount to defining functions where the order of pattern matching has an effect?  But, if not `ForkL` and `JoinL`, what _should_ be the way to pattern match linear maps?

My conclusion is that we should not match on the constructor that happens to be at the top level.  Rather, our "patterns" should simply be matches on cardinality.  For instance, imagine a pattern with an `n + 1` type that matches on any input whose cardinality is greater than 1.  It would yield two smaller values, one with cardinality `1` (e.g., a `Par1`) and the other whose cardinality is `n`.  This type of pattern is the natural conclusion to the idea of failable patterns I propose in this PR.

This brings me back to my last sentence in the PR description: is this a good thing?  Should we strive for these "cardinality patterns"?  The trade-off is in _control_.  Imagine two functions `fooCon` and `fooCard` where `fooCon` pattern matches on _con_-structors and `fooCard` pattern matches on _card_-inality.  A user of `fooCon` can provide data in different forms and potentially see different performance results.  For instance, if she passes in a balanced, parallelizable data structure, she may see good parallelism and not otherwise.  A user of `fooCard` will always see the same performance results regardless of how she structures her data.  To be fair, `fooCon` is strictly more general than `fooCard` because we could define `fooCard = fooCon . iso` where `iso` is an isomorphism from whatever data was provided to a canonicalized form, but is that generality something we even want?  Can we write better functions (for some definition of "better") by matching on cardinality than by performing an isomorphism and then matching on the constructors?  More generally, are linear maps defined by structure itself or by cardinality of structure?

bolt12:
That's an interesting take but I can not see how that would be possible and even if it would it seems to me that pattern matching on cardinality seems like an implementation detail to overcome the type checker limitation. It feels more natural to pattern match on Joins and Forks to me, but I agree that in an hipothetical `fooCard` parallel performance would be more predicable, but at what cost?

 It sure looks interesting though, but it makes me question performance issues as well.

dwincort:
> That's an interesting take but I can not see how that would be possible

I haven't figured out how to use `PatternSynonyms` to do this, but I think the idea is sound.  You could have something like:
```haskell
foo :: (CardOf f ~ n, CardOf g ~ m) => L f g a -> ...
foo
  | natVal n == 1 = ...
  | even (natVal m) = ...
  | otherwise = ...
```
(Here, `CardOf` is a type family that returns a `Nat` representing the cardinality of the representation of the given type.)  You'd also need appropriate isomorphisms, but it's certainly possible to construct:
```haskell
isoCard :: (V4 f f' g g', CardOf f ~ CardOf f', CardOf g ~ CardOf g') => L f g a -> L f' g' a
```
(Well, I've written such a thing assuming that the cardinalities are finite.)

---

> and even if it would it seems to me that pattern matching on cardinality seems like an implementation detail to overcome the type checker limitation.

One thing that I keep thinking about is [Strassen's algorithm for matrix multiplication](https://en.wikipedia.org/wiki/Strassen_algorithm).  I feel like a system for doing linear algebra should, at some level, be able to implement this algorithm, but the system we have falls short.  In fact, this whole effort of failable patterns was an attempt to overcome this -- I was hoping that I could find a set of patterns where strassen was implementable -- and in that sense, I succeeded, as with these failable patterns, it's possible to write a form of strassen.  Wooh!

That said, it's not perfect.  Imagine trying to multiply two 7x7 matrices.  It's possible that the most efficient option is to use Strassen, but to do so, you must first pad your matrices with zeros so that they're 8x8.  We can argue about the meaning of "pad with zeros", but this does have a translation to linear maps, that of lifting the mapping into a higher dimensional space.  As far as practice, in the current system, the only way to perform this lifting would be to use binary forks and joins to add rows/columns of zeros.  However, now Strassen is impossible because our dimensions of cardinality 8 are actually represented as dimensions of cardinality `(7 + 1)` instead of `(2 * 4)`.  The failable patterns I've built won't work.

This was a very long-winded way of saying that this is not just an implementation detail or an effort to get around type checking but a fundamental difference in the level of abstraction we are allowed to use when writing operations.  Furthermore, with Strassen as an example, this difference can have a very real impact on performance.

dwincort:
For what it's worth, there is a feature request on GHC to make better `complete` signatures, so I won't put that together:
https://gitlab.haskell.org/ghc/ghc/-/issues/16892

I don't see anything about the weird performance issue with the complete pragma, so if I get some time, I'll try to make a minimal working example and submit a bug report.

conal:
> We already accept that if we have a fork of joins then that should be equivalent to a join of forks. Indeed, the pattern synonyms @conal already has do this. But, what if we have a join of joins — can we reorder those joins? I feel like we should be able to (and I don’t think it matters if the joins are both binary, both n-ary, or one of each).

I accept the fork/join equivalence because it’s a valid theorem about linear maps, without regard to representation, i.e., because it *is* true, not because it should be.

More generally, this “exchange law” is true for all cocartesian categories (not just biproduct categories). See [*Calculating Functional Programs*](https://www.cs.ox.ac.uk/publications/publication2360-abstract.html), section 1.5.4.

Is there an analogous truth about reordering joins? Would such a claim even type-check?

> Can we do even more? What if we have a binary join of two things that happen to be the same — can we change that into an n-ary join where our representable functor has cardinality two? This question is the big one and really gets to the heart of my wrestling. **If we think of the things we’re manipulating as matrices**, then easily, the answer should be yes. But should it be true for linear maps in general? I can’t come up with a counterexample, but that may be a failure of imagination on my part.

(Emphasis added.)

Great questions\! I know what my answer is: this project is *not* about matrices, because linear algebra itself is not about matrices. Instead, the project is about a better conceptual model and programming interface to linear maps than matrices provide.

An efficient implementation will probably use a representation other than `L`. One such implementation might use a conventional 2D array matrix representation, but probably really a 1D array, since 2D arrays are an abstraction over 1D. In fact, 1D arrays are abstractions over the perfect binary leaf trees embedded in memory chips. (Such trees are the natural trie when indices are bit strings, with address look-up being trie indexing.) Even this last statement harmfully assumes the processor/memory von Neumann bottleneck, which is one of the fundamental sources of inefficiency of CPUs. IIUC, this bottleneck accounts for most time and power expended by CPUs. I want to take advantage of very efficient architectures that distribute memory so that it’s closer to various computations.

Other efficient implementations make great advantage of abandoning matrices altogether, as in my simple, efficient, parallel-friendly, correct-by-construction, reverse-mode gradient algorithm (better “backpropagation”). I suspect that the AD and ML communities missed this algorithm for decades largely because they think of linear algebra in terms of matrices and thus mis-defined the problems they were trying to solve. (The other fundamental mistake was to think in terms of graphs.)

Every matrix-oriented API (all current “linear algebra” libraries I know of) thus harms modularity and safety of programming *and* efficiency of modern (non-CPU) implementation.

I do intend to introduce arrays and cardinality into *implementations*, but not into the programming model and hence not in the API (except for conversion of externally generated data). This part of the story is in [*Data-Parallel Programming without Arrays*](http://conal.net/papers/data-parallel-programming-without-arrays).

conal:
> One thing that I keep thinking about is [Strassen’s algorithm for matrix multiplication](https://en.wikipedia.org/wiki/Strassen_algorithm). \[…\]

I’m glad\! Being able to use the Strassen algorithm for linear map composition interests me greatly, and I’m confident that we can do so while staying entirely faithful to the denotational model (linear maps).

dwincort:
> Great questions! I know what my answer is: this project is _not_ about matrices, because linear algebra itself is not about matrices. Instead, the project is about a better conceptual model and programming interface to linear maps than matrices provide.

I get that!  Hence the critical question of my paragraph: "But should it be true for linear maps in general?"  I completely get that just because something is true for a matrix doesn't mean it's true for a linear map, and also I am deeply thinking about what it means for something to be a linear map, what the _essence_ of a linear map is.  Maybe I'll try to avoid even using the word matrix in the future.  Let me try again (after a bit of a dive on wikipedia).

The idea of isomorphisms over cardinality only holds when a cardinality exists.  In this sense, I'm probably limiting my view of linear maps to acting over free modules whose underlying ring satisfies the invariant basis number condition.  I say "probably" because this limit may be overly strict; that said, all free modules I describe are isomorphic to any other modules with the same cardinality.  Now, I know we're requiring the `f` and `g` of our `L f g s` type to be `Representable`, and as far as I can tell, any `Representable` type has a cardinality.  Doesn't that mean we're already limiting ourselves to a certain class of linear maps, ones that are over representable modules?  And if I'm right that representability implies cardinality, then given we're already requiring `Representable`, why not use the cardinality therein?

I myself can come up with two good answers.  One, we should think of `Representable` as an annoying implementation detail and be diligent about pursuing linear maps with _no_ restrictions whatsoever.  Two, I'm wrong about cardinality being a feature of representability.  (In which case, my imagination is failing me: what's an example of a a `Representable` functor that does not have an exact cardinality?)

I guess one of my fundamental questions is: how much are we losing if we restrict ourselves to free modules over rings satisfying the invariant basis number condition?  It's kind of related to: how much are we gaining my staying diligent to using linear maps?  I don't _think_ we're losing that much, but I want to know exactly what it is.

I may need to spend a bit more time on wikipedia (or in an algebra textbook) to get some answers.

---

> I do intend to introduce arrays and cardinality into implementations, but not into the programming model and hence not in the API (except for conversion of externally generated data).

This seems totally acceptable to me, but with one caveat: the implementation of composition.  Right now, we have an implementation of matrix composition in terms of `L`.  Is this necessary?  Would you intend to use this in production code (perhaps with components of it optimized or rewritten, but the general form would stay the same)?  If so, I feel like we're not done making a general, performant algorithm.  If not, I guess I'm confused about the point of some of our efforts here -- it seems like all we'd need to write are the classes and type signatures, and we could omit all of the implementations.

---

> We could try translating an array-oriented implementation of Strassen, but I’d expect better results from translating a correctness proof, where the essential insights are to be found.

I'm not sure I follow this.  In some sense, Strassen is a very simple algorithm that works in very specific cases (when the input mappings can be quartered).  As far as I was aware, the proof is simply an exercise in basic algebra (distributive and commutative properties, etc.) and using the product and coproduct laws along with the fact that the quarters are properly composable.  Is there some more fundamental proof that you're thinking of?

dwincort:
> Is there an analogous truth about reordering joins? Would such a claim even type-check?

I mean, we can write:
```haskell
unjoinjoin1 :: V4 f g h k => L ((h :.: f) :*: (h :.: g)) k s -> L (h :.: (f :*: g)) k s
unjoinjoin1 ((Join p) :| (Join q)) = Join $ liftR2 (:|) p q

unjoinjoin2 :: V4 f g h k => L (h :.: (f :*: g)) k s -> L ((h :.: f) :*: (h :.: g)) k s
unjoinjoin2 (Join (unzip . fmap unjoin2 -> (p,q))) = Join p :| Join q

unjoinjoin3 :: V5 f g h j k => L ((f :*: g) :*: (h :*: j)) k s -> L ((f :*: h) :*: (g :*: j)) k s
unjoinjoin3 ((f :| g) :| (h :| j)) = (f :| h) :| (g :| j)

unjoinjoin4 :: V4 f g h k => L (f :.: (g :.: h)) k s -> L (g :.: (f :.: h)) k s
unjoinjoin4 (Join (fmap unjoinL -> x)) = Join $ fmap Join $ distribute x
```
I don't know how to join (pun intended) these functions into a single function, but yeah, we can reorder joins of joins.

dwincort:
We can also do the second thing I suggested, converting a binary join where the two parts happen to be the same into an n-ary join with a representable functor that happens to have cardinality 2:
```haskell
joinToPair :: V2 f g => L (f :*: f) g s -> L (Pair :.: f) g s
joinToPair (p :| q) = Join (p :# q)
```
This particular implementation chooses the functor to be `Pair`, but if you _know_ the cardinality of the functor and leverage representability, you can rewrite this to take the new functor as a type parameter.

dwincort:
FYI, I've opened up a new PR for failable patterns that is _much_ cleaner than this one: https://github.com/conal/linalg/pull/32.

Also, I've learned a bit more about linear maps, and I'm going to try to keep digging into the math today.  It's only a _little_ over my head ;).

I'm wondering if I'm too tied to the idea of `Representable`.  Are we just using it as a convenient implementation tool, or are we making a statement about what kind of linear maps we want to support, i.e., ones that are made up of representable functors?  For instance, I can write definitions for `unjoinjoin` and `joinToPair` because we have `Representable`, but are those isomorphisms that should hold for linear maps in general?  I'm now trying to figure that out.

conal:
, however> > Is there an analogous truth about reordering joins? Would such a claim even type-check?
> 
> I mean, we can write: \[…\]
> 
> I don’t know how to join (pun intended) these functions into a single function, but yeah, we can reorder joins of joins.

Thanks for these examples.

There are a lot of *transformations* on linear maps, and many of those transformations are *isomorphisms*, but the join/fork exchange laws are *equalities*. If we phrased your transformation examples as claims about equality, those claims would not type-check.

PR #28, opened 2020-07-28
Specification

conal:
This PR is work in progress, but I want to share what I’m up to.

I started a new module (currently called “`F`”) with a definition of linear functions as part of a denotational specification of `L`. I want to use a shared vocabulary, so I’m also working on some category classes that are poly-kinded (to accommodate the functor categories we’ll need) and have associated constraints, products, and coproducts. Going okay so far, but we’ll see how type inference works out.

The specification will include a “semantic function” mapping our `L` type to linear functions, plus the requirement that this semantic function is homomorphic for *all* of the vocabulary we use. I expect this discipline to give us a precise, regular, and intuitively helpful specification of all we do, with which we can calculate and/or prove our implementation code.

Still to come:

  - Adapt `L` to use this vocabulary.
  - Solve a typing puzzle with n-ary products and coproducts (noted in code).
  - Rename modules, maybe adding a module name-space level (e.g., “`LinAlg.L`”).
  - Define the semantic function, which should be very pretty, due to the homomorphisms and the close relationship of the `L` representation with the vocabulary.
  - Demonstrate some code calculations.
  - `Representable` instances for linear functions and for `L`. As with all other interfaces, the semantic function for `L` must be a `Representable` homomorphism. See \#26 for discussion.

I won’t merge until at least the first two steps are done.

Comments & questions most welcome\!

### DIFF
--- a/src/Category.hs
+++ b/src/Category.hs
@@ -10,12 +10,13 @@ import qualified Prelude as P
 import Prelude hiding (id,(.))
 import GHC.Types (Constraint)
 import qualified Control.Arrow as A
+import GHC.Generics ((:.:)(..))
 import Data.Functor.Rep
 
 import Misc
 
-class    Yes1 a
-instance Yes1 a
+class    Unconstrained a
+instance Unconstrained a
 
 class Category (k :: u -> u -> *) where
   type Obj k :: u -> Constraint
@@ -68,7 +69,7 @@ pattern f :& g <- (unfork2 -> (f,g)) where (:&) = (&&&)
 --   A type signature must be provided for a set of polymorphic pattern synonyms.
 --   In {-# complete :& #-}
 --
--- Given the generality of (:&), is there any solution to this completeness issue?
+-- Instead, give a typed COMPLETE pragma with each cartesian category instance.
 
 class Category k => Comonoidal k co | k -> co where
   infixr 2 +++
@@ -98,36 +99,22 @@ pattern f :| g <- (unjoin2 -> (f,g)) where (:|) = (|||)
 -- {-# complete (:|) #-}  -- See (:&) above
 
 
--- -- N-ary (representable) counterparts
+-- -- N-ary (representable) counterparts.
+
+-- Assumes functor categories. To do: look for a clean, poly-kinded alternative.
+-- I guess we could generalize from functor composition and functor application.
 
 class (Category k, Representable r) => MonoidalR k r where
-  cross :: r (a `k` b) -> (r a `k` r b)
-
--- TODO: maybe "N" (for n-ary) instead of "R" (for representable) in these
--- class names. On the other hand, people will probably assume n is a number.
-
--- TODO: keep Representable r superclass constraint?
-
--- Oops! The type r (a `k` b) requires r :: * -> *, hence r a :: *, a :: *.
--- The Representable constraint does as well.
---
---   λ> :k Monoidal
---   Monoidal :: (u -> u -> *) -> (u -> u -> u) -> Constraint
---   λ> :k MonoidalR
---   MonoidalR :: (* -> * -> *) -> (* -> *) -> Constraint
---
--- For L, we have
--- 
---   cross :: (...) => c (L a b s) -> L (c :.: a) (c :.: b) s
---
--- What generalization are we after here?
+  cross :: r (a `k` b) -> ((r :.: a) `k` (r :.: b))
 
 class MonoidalR k r => CartesianR k r where
-  exs :: r (r a `k` a)
-  dups :: a `k` r a
+  exs :: r ((r :.: a) `k` a)
+  dups :: a `k` (r :.: a)
 
-fork :: (CartesianR k r, Obj2 k a c) => r (a `k` c) -> (a `k` r c)
+fork :: (CartesianR k r, Obj2 k a c) => r (a `k` c) -> (a `k` (r :.: c))
 fork fs = cross fs . dups
+
+#if 0
 
 class (Category k, Representable r) => ComonoidalR k r where
   plus :: r (a `k` b) -> ((Rep r :* a) `k` (Rep r :* b))
@@ -148,10 +135,17 @@ class (Representable r, ComonoidalR k r) => CocartesianR k r where
 -- In the ambiguity check for ‘jams’
 -- To defer the ambiguity check to use sites, enable AllowAmbiguousTypes
 
+#endif
+
+-- Try specializing to bifunctor categories instead. Reconsider naming.
+class CartesianR k r => BifunctorR k r where
+  ins :: r (a `k` (r :.: a))
+  jams :: (r :.: a) `k` a
+
 -- -- Instances
 
 instance Category (->) where
-  type Obj (->) = Yes1
+  type Obj (->) = Unconstrained
   id = P.id
   (.) = (P..)
 
@@ -174,6 +168,10 @@ instance Cocartesian (->) (:+) where
   -- jam (Left  a) = a
   -- jam (Right a) = a
 
+#if 0
+
+-- These instances are broken now that MonoidalR etc assumes functor categories.
+
 instance Representable p => MonoidalR (->) p where
   cross = liftR2 ($)
 
@@ -187,3 +185,5 @@ instance Representable p => ComonoidalR (->) p where
 instance Representable p => CocartesianR (->) p where
   ins = tabulate (,)  -- = tabulate (\ i a -> (i, a))
   jams = snd
+
+#endif
--- a/src/Category.hs
+++ b/src/Category.hs
@@ -1,4 +1,8 @@
--- {-# OPTIONS_GHC -Wno-unused-imports #-} -- TEMP
+{-# LANGUAGE QuantifiedConstraints #-}
+{-# LANGUAGE UndecidableInstances #-} -- see below
+{-# LANGUAGE UndecidableSuperClasses #-} -- see below
+
+{-# OPTIONS_GHC -Wno-unused-imports #-} -- TEMP
 
 {-# LANGUAGE AllowAmbiguousTypes #-}   -- See below
 
@@ -18,13 +22,27 @@ import Misc
 class    Unconstrained a
 instance Unconstrained a
 
+-- https://github.com/conal/linalg/pull/28#issuecomment-670313952
+class    Obj' k a => Obj k a
+instance Obj' k a => Obj k a
+
+-- Illegal constraint ‘Obj' k a’ in a superclass context
+--   (Use UndecidableInstances to permit this)
+
+-- Potential superclass cycle for ‘Obj’
+--   one of whose superclass constraints is headed by a type family:
+--     ‘Obj' k a’
+-- Use UndecidableSuperClasses to accept this
+
 class Category (k :: u -> u -> *) where
-  type Obj k :: u -> Constraint
+  type Obj' k :: u -> Constraint
   infixr 9 .
   id :: Obj k a => a `k` a
-  (.) :: (b `k` c) -> (a `k` b) -> (a `k` c)
+  (.) :: Obj3 k a b c => (b `k` c) -> (a `k` b) -> (a `k` c)
 
--- Experiment: omit Obj constraints for argument arrows.
+-- TODO: does (.) really need these constraints? We may know better when we try
+-- "matrices" (non-inductive and inductive) with and without these (.)
+-- constraints. Similarly for other classes.
 
 type Obj2 k a b         = C2 (Obj k) a b
 type Obj3 k a b c       = C3 (Obj k) a b c
@@ -35,9 +53,13 @@ type Obj6 k a b c d e f = C6 (Obj k) a b c d e f
 -- TODO: Maybe eliminate all type definitions based on Obj2 .. Obj6 in favor of
 -- their definitions, which are not much longer anyway.
 
-class Category k => Monoidal k p | k -> p where
+-- Products of objects are objects.
+-- Seee https://github.com/conal/linalg/pull/28#issuecomment-670313952
+type ObjBin k p = ((forall a b. Obj2 k a b => Obj k (a `p` b)) :: Constraint)
+
+class (Category k, ObjBin k p) => Monoidal k p | k -> p where
   infixr 3 ***
-  (***) :: (a `k` c) -> (b `k` d) -> ((a `p` b) `k` (c `p` d))
+  (***) :: Obj4 k a b c d => (a `k` c) -> (b `k` d) -> ((a `p` b) `k` (c `p` d))
 
 -- The functional dependency requires p to be uniquely determined by k. Might
 -- help type inference. Necessitates a "Comonoidal" class with "(+++)", which is
@@ -46,6 +68,22 @@ class Category k => Monoidal k p | k -> p where
 
 -- TODO: make p an associated type, and see how the class and instance
 -- definitions look in comparison.
+--
+-- @dwincort (https://github.com/conal/linalg/pull/28#discussion_r466989563):
+-- From what I can tell, if we use `QuantifiedConstraints` with `p`, then we
+-- can't turn it into an associated type. I'm not sure that's so bad, but it's
+-- worth noting.
+--
+-- TODO: keep poking at this question.
+
+-- TODO: Does it make any sense to move 'p' and its ObjBin into the method
+-- signatures, as in MonoidalR below? Should we instead move 'r' in MonoidalR
+-- from the method signatures to the class? It feels wrong to me (conal) that
+-- there is only one binary product but many n-ary. In other sense, n-ary is
+-- even more restrictive than binary: the (type-indexed) tuple-ness of
+-- representable functors is wired in, and so is the object kind. For instance,
+-- we cannot currently handle n-ary coproducts that are not n-ary cartesian
+-- *products*.
 
 class Monoidal k p => Cartesian k p where
   exl :: Obj2 k a b => (a `p` b) `k` a
@@ -58,10 +96,8 @@ infixr 3 &&&
       => (a `k` c) -> (a `k` d) -> (a `k` (c `p` d))
 f &&& g = (f *** g) . dup
 
--- Can I instead extract the Obj constraints from f and g?
-
 -- Inverse of uncurry (&&&)
-unfork2 :: (Cartesian k p, Obj2 k c d)
+unfork2 :: (Cartesian k p, Obj3 k a c d)
         => (a `k` (c `p` d)) -> ((a `k` c) :* (a `k` d))
 unfork2 f = (exl . f , exr . f)
 
@@ -79,9 +115,9 @@ pattern f :& g <- (unfork2 -> (f,g)) where (:&) = (&&&)
 --
 -- Instead, give a typed COMPLETE pragma with each cartesian category instance.
 
-class Category k => Comonoidal k co | k -> co where
+class (Category k, ObjBin k co) => Comonoidal k co | k -> co where
   infixr 2 +++
-  (+++) :: (a `k` c) -> (b `k` d) -> ((a `co` b) `k` (c `co` d))
+  (+++) :: Obj4 k a b c d => (a `k` c) -> (b `k` d) -> ((a `co` b) `k` (c `co` d))
 
 -- TODO: keep both Monoidal and Comonoidal or have one class with two instances
 -- per category?
@@ -98,7 +134,7 @@ infixr 2 |||
 f ||| g = jam . (f +++ g)
 
 -- Inverse of uncurry (|||)
-unjoin2 :: (Cocartesian k co, Obj2 k a b)
+unjoin2 :: (Cocartesian k co, Obj3 k a b c)
         => ((a `co` b) `k` c) -> ((a `k` c) :* (b `k` c))
 unjoin2 f = (f . inl , f . inr)
 
@@ -118,7 +154,7 @@ class (Cartesian k p, Cocartesian k p) => Biproduct k p
 -- -- Instances
 
 instance Category (->) where
-  type Obj (->) = Unconstrained
+  type Obj' (->) = Unconstrained
   id = P.id
   (.) = (P..)
 
@@ -147,21 +183,37 @@ instance Cocartesian (->) (:+) where
 -- Assumes functor categories. To do: look for a clean, poly-kinded alternative.
 -- I guess we could generalize from functor composition and functor application.
 
+-- N-ary products of objects are objects.
+
+-- type ObjR k r = (((forall a. Obj k a => Obj k (r :.: a)) :: Constraint)
+--                 , Representable r)
+
+-- Illegal polymorphic type:
+--   forall (a :: k1 -> *). Obj k a => Obj k (r :.: a)
+-- GHC doesn't yet support impredicative polymorphism
+
+type ObjR' k r = ((forall a. Obj k a => Obj k (r :.: a)) :: Constraint)
+
+class    (Representable r, ObjR' k r) => ObjR k r
+instance (Representable r, ObjR' k r) => ObjR k r
+
 class Category k => MonoidalR k where
-  cross :: Representable r => r (a `k` b) -> ((r :.: a) `k` (r :.: b))
+  cross :: (Obj2 k a b, ObjR k r)
+        => r (a `k` b) -> ((r :.: a) `k` (r :.: b))
 
 -- TODO: maybe wire in p = (:*:) for Monoidal, since we're doing essentially the
 -- same for MonoidalR by choosing Representable r and (:.:).
 
 class MonoidalR k => CartesianR k where
-  exs :: Representable r => r ((r :.: a) `k` a)
-  dups :: Representable r => a `k` (r :.: a)
+  exs :: (Obj k a, Representable r) => r ((r :.: a) `k` a)
+  dups :: (Obj k a, Representable r) => a `k` (r :.: a)
 
-fork :: (CartesianR k, Obj2 k a c, Representable r)
+fork :: (CartesianR k, Obj2 k a c, ObjR k r)
      => r (a `k` c) -> (a `k` (r :.: c))
 fork fs = cross fs . dups
 
-unfork :: (CartesianR k, Representable r) => a `k` (r :.: b) -> r (a `k` b)
+unfork :: (CartesianR k, Obj2 k a b, ObjR k r)
+       => a `k` (r :.: b) -> r (a `k` b)
 unfork f = (. f) <$> exs
 
 -- TODO: How can we know that fork and unfork form an isomorphism?
@@ -174,11 +226,11 @@ class CartesianR k => BiproductR k where
 -- TODO: Maybe replace (Representable r, Eq (Rep r), Foldable r) with an
 -- associated functor constraint.
 
-join :: (BiproductR k, Representable r, Foldable r, Obj2 k a b)
+join :: (BiproductR k, ObjR k r, Foldable r, Obj2 k a b)
      => r (a `k` b) -> (r :.: a) `k` b
 join fs = jams . cross fs  -- note cross == plus
 
-unjoin :: (BiproductR k, Obj2 k a b, Representable r, Eq (Rep r))
+unjoin :: (BiproductR k, Obj2 k a b, ObjR k r, Eq (Rep r))
        => (r :.: a) `k` b -> r (a `k` b)
 unjoin f = (f .) <$> ins
 
@@ -186,7 +238,4 @@ unjoin f = (f .) <$> ins
 -- defaults and giving defaults for exs and dups in terms of fork and unfork.
 -- Ditto for ins/jams and join/unjoin. Use MINIMAL pragmas.
 
--- TODO: Abelian
-
-class (Biproduct (l s) p, BiproductR (l s)) => Scalable s l p where
-  scale :: s -> l s Par1 Par1
+-- Add Abelian?
--- a/src/Category.hs
+++ b/src/Category.hs
@@ -190,5 +190,4 @@ unjoin f = (f .) <$> ins
 
 class (Biproduct (l s) p, BiproductR (l s)) => Linear s l p where
   scale :: s -> l s Par1 Par1
-  infixl 9 @@  -- Linear application
-  (@@) :: l s a b -> a s -> b s  -- linear
+  at :: l s a b -> a s -> b s  -- linear
--- a/src/Category.hs
+++ b/src/Category.hs
@@ -70,9 +70,9 @@ class (Category k, ObjBin k p) => Monoidal k p | k -> p where
 -- definitions look in comparison.
 --
 -- @dwincort (https://github.com/conal/linalg/pull/28#discussion_r466989563):
--- From what I can tell, if we use `QuantifiedConstraints` with `p`, then we
+-- "From what I can tell, if we use `QuantifiedConstraints` with `p`, then we
 -- can't turn it into an associated type. I'm not sure that's so bad, but it's
--- worth noting.
+-- worth noting." See also the GHC error message there.
 --
 -- TODO: keep poking at this question.
 
@@ -101,7 +101,7 @@ unfork2 :: (Cartesian k p, Obj3 k a c d)
         => (a `k` (c `p` d)) -> ((a `k` c) :* (a `k` d))
 unfork2 f = (exl . f , exr . f)
 
--- TODO: How can we know that uncurry (&&&) and unfork2 form an isomorphism?
+-- Exercise: Prove that uncurry (&&&) and unfork2 form an isomorphism.
 
 pattern (:&) :: (Cartesian k p, Obj3 k a c d)
              => (a `k` c) -> (a `k` d) -> (a `k` (c `p` d))
@@ -138,7 +138,7 @@ unjoin2 :: (Cocartesian k co, Obj3 k a b c)
         => ((a `co` b) `k` c) -> ((a `k` c) :* (b `k` c))
 unjoin2 f = (f . inl , f . inr)
 
--- TODO: How can we know that uncurry (|||) and unjoin2 form an isomorphism?
+-- Exercise: Prove that uncurry (|||) and unjoin2 form an isomorphism.
 
 pattern (:|) :: (Cocartesian k co, Obj3 k a b c)
              => (a `k` c) -> (b `k` c) -> ((a `co` b) `k` c)
@@ -191,7 +191,7 @@ unfork :: (CartesianR k, Obj2 k a b, ObjR k r)
        => a `k` (r :.: b) -> r (a `k` b)
 unfork f = (. f) <$> exs
 
--- TODO: How can we know that fork and unfork form an isomorphism?
+-- Exercise: Prove that fork and unfork form an isomorphism.
 
 -- N-ary biproducts
 class CartesianR k => BiproductR k where
--- a/src/Category.hs
+++ b/src/Category.hs
@@ -10,7 +10,6 @@ import qualified Prelude as P
 import Prelude hiding (id,(.))
 import GHC.Types (Constraint)
 import qualified Control.Arrow as A
--- import qualified Data.Tuple as T
 import Data.Functor.Rep
 
 import Misc
@@ -36,21 +35,40 @@ class Category k => Monoidal k p | k -> p where
   infixr 3 ***
   (***) :: (a `k` c) -> (b `k` d) -> ((a `p` b) `k` (c `p` d))
 
--- Should we require p to be uniquely determined by k? Might help type
--- inference. We'd then need a "Comonoidal" class with "(+++)", which is
--- perhaps better.
+-- The functional dependency requires p to be uniquely determined by k. Might
+-- help type inference. Necessitates a "Comonoidal" class with "(+++)", which is
+-- perhaps better than giving two Monoidal instances for a single category (eg
+-- for (->)).
 
 class Monoidal k p => Cartesian k p where
   exl :: Obj2 k a b => (a `p` b) `k` a
   exr :: Obj2 k a b => (a `p` b) `k` b
   dup :: Obj  k a   => a `k` (a `p` a)
 
+-- Binary fork
 infixr 3 &&&
 (&&&) :: (Cartesian k p, Obj3 k a c d)
       => (a `k` c) -> (a `k` d) -> (a `k` (c `p` d))
 f &&& g = (f *** g) . dup
 
 -- Can I instead extract the Obj constraints from f and g?
+
+-- Inverse of uncurry (&&&)
+unfork2 :: (Cartesian k p, Obj2 k c d)
+        => (a `k` (c `p` d)) -> ((a `k` c) :* (a `k` d))
+unfork2 f = (exl . f , exr . f)
+
+pattern (:&) :: (Cartesian k p, Obj3 k a c d)
+             => (a `k` c) -> (a `k` d) -> (a `k` (c `p` d))
+pattern f :& g <- (unfork2 -> (f,g)) where (:&) = (&&&)
+-- {-# complete (:&) #-}
+
+-- GHC error:
+-- 
+--   A type signature must be provided for a set of polymorphic pattern synonyms.
+--   In {-# complete :& #-}
+--
+-- Given the generality of (:&), is there any solution to this completeness issue?
 
 class Category k => Comonoidal k co | k -> co where
   infixr 2 +++
@@ -64,10 +82,20 @@ class Comonoidal k co => Cocartesian k co where
   inr :: Obj2 k a b => b `k` (a `co` b)
   jam :: Obj  k a   => (a `co` a) `k` a
 
+-- Binary join
 infixr 2 |||
 (|||) :: (Cocartesian k co, Obj3 k a b c)
       => (a `k` c) -> (b `k` c) -> ((a `co` b) `k` c)
 f ||| g = jam . (f +++ g)
+
+-- Inverse of uncurry (|||)
+unjoin2 :: (Cocartesian k co, Obj2 k a b) => ((a `co` b) `k` c) -> ((a `k` c) :* (b `k` c))
+unjoin2 f = (f . inl , f . inr)
+
+pattern (:|) :: (Cocartesian k co, Obj3 k a b c)
+             => (a `k` c) -> (b `k` c) -> ((a `co` b) `k` c)
+pattern f :| g <- (unjoin2 -> (f,g)) where (:|) = (|||)
+-- {-# complete (:|) #-}  -- See (:&) above
 
 
 -- -- N-ary (representable) counterparts
--- a/src/Category.hs
+++ b/src/Category.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE QuantifiedConstraints #-}
 {-# LANGUAGE UndecidableInstances #-} -- see below
 {-# LANGUAGE UndecidableSuperClasses #-} -- see below
 
--- a/src/Category.hs
+++ b/src/Category.hs
@@ -175,46 +175,42 @@ class (Cartesian k p, Cocartesian k p) => Biproduct k p
 --   forall (a :: k1 -> *). Obj k a => Obj k (r :.: a)
 -- GHC doesn't yet support impredicative polymorphism
 
-type ObjR' k r = ((forall a. Obj k a => Obj k (r :.: a)) :: Constraint)
+type ObjR' k r = ((forall z. Obj k z => Obj k (r :.: z)) :: Constraint)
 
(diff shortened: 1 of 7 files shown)


src/Category.hs
```diff
@@ -0,0 +1,110 @@
...
+class Monoidal k c => Cocartesian k c where
+  inl :: Obj2 k a b => a `k` (a `c` b)
+  inr :: Obj2 k a b => b `k` (a `c` b)
+  jam :: Obj  k a   => (a `c` a) `k` a
+
+-- N-ary (representable) counterparts
+
+class Representable p => MonoidalRep k p where
+  exs :: p (p a `k` a)
+  dups :: a `k` p a
```
bolt12, line 62: Should this be:

```haskell
class Representable p => MonoidalRep k p where
  exs :: Representable (m p) => p (p `m` a `k` a) -- p (F s (p :.: a) a | p ((p :.: a) -> a)
  dups :: Representable (m p) => a `k` (p `m` a)
```

I'm just not sure how to implement 

```haskell
instance Representable p => MonoidalRep (->) p where
  exs = _ -- tabulate (flip index)
  dups = pureRep
```

bolt12: Also shouldn't the class be called `CartesianRep` ?

conal: Yes; thanks. I’ve renamed to `CartesianRep`, then to `CartesianR`, added `MonoidalR` with `cross`, and corresponding coproduct classes and instances.

src/Misc.hs
```diff
@@ -0,0 +1,72 @@
+-- {-# OPTIONS_GHC -Wno-unused-imports #-} -- TEMP
+
+-- | Miscellany
+
+module Misc where
+
+import qualified Prelude as P
+import Prelude hiding ((+),sum,(*),unzip)
```
dwincort, line 8: For the sake of elegance, maybe instead of calling this `Misc`, we should make this file our own `Prelude` (`LinAlgPrelude`?).  Then, we can turn on `NoImplicitPrelude` in the cabal file and just always import this module.

Not only does this mean we can stop with this `import Prelude hiding` business, but we can also have our custom Prelude export any number of things that we use all the time by default.  For instance, we can have it export `Constraint`, the `Generic` types we care about, and even the whole `Data.Functor.Rep` and `Data.Distributive` modules.

We could go much further if we wanted, like excising all the partial functions (e.g., perhaps starting with [Protolude](https://hackage.haskell.org/package/protolude)) — it's up to us.

src/Misc.hs
```diff
@@ -0,0 +1,75 @@
...
+  (*) :: a -> a -> a
+
+instance Additive Double where { zero = 0; (+) = (P.+) }
+instance Semiring Double where { one  = 1; (*) = (P.*) }
+
+sum :: (Foldable f, Additive a) => f a -> a
+sum = foldr (+) zero
+
+-- To do: maybe replace `foldr` by `foldl` or `foldl'`. Does a strict left fold
+-- help at all, considering that our "vectors" aren't flat?
```
dwincort, line 61: Another option is to recreate the `Sum` newtype from `Data.Monoid` but using `Additive` instead of `Num`.  Then, you can write:
```haskell
sum = getSum . foldMap Sum
```
(or maybe something more efficient with `coerce`?).

src/Category.hs
conal: > Should this be:
> 
> ``` haskell
> class Representable p => MonoidalRep k p where
>   exs :: Representable (m p) => p (p `m` a `k` a) -- p (F s (p :.: a) a | p ((p :.: a) -> a)
>   dups :: Representable (m p) => a `k` (p `m` a)
> ```

Interesting\! Noodling through the types, `Representable p` and `Representable (m p)` imply `m :: (* -> *) -> (* -> *) -> (* -> *)`, and ``p `m` a`` then implies `a :: * -> *`. We couldn’t then accommodate `k :: * -> * -> *`, `a :: *` (as in the `(->)` category).

Now suppose `k :: u -> u -> *`, as in the general case of `Category`. Then `a :: u`, `m p a :: u`, and so `m :: (* -> *) -> u -> u`. When `u = * -> *` (as in linear maps & its representations), `m` is functor composition, i.e., `m = (:.:)`. When `u = *` (as in ordinary functions), `m` is instead functor *application*. Since Haskell type inference does only *first-order* unification, `m` cannot really be instantiated to type application. We could instead make a `newtype` wrapper:

``` haskell
newtype Ap f a :: Ap (f a)
```

I guess the result could be usable though not convenient, since we’d have to manage the `Ap` wrappers.

src/Category.hs
```diff
@@ -0,0 +1,189 @@
...
+  exr :: Obj2 k a b => (a `p` b) `k` b
+  dup :: Obj  k a   => a `k` (a `p` a)
+
+-- Binary fork
+infixr 3 &&&
+(&&&) :: (Cartesian k p, Obj3 k a c d)
+      => (a `k` c) -> (a `k` d) -> (a `k` (c `p` d))
+f &&& g = (f *** g) . dup
+
+-- Can I instead extract the Obj constraints from f and g?
```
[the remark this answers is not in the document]
conal: In [concat](https://github.com/conal/concat) (“constrained categories”, but really compiling-to-categories), I included the explicit `Obj` constraints (which I called “`Ok`”) in methods and utility functions, but I can imagine getting them instead from argument arrows like `f` and `g` by adding a method to `Category` (which introduces `Obj`):

``` haskell
class Category (k :: u -> u -> *) where
  type Obj k :: u -> Constraint
  infixr 9 .
  id :: Obj k a => a `k` a
  (.) :: (b `k` c) -> (a `k` b) -> (a `k` c)
  isObj :: (a `k` b) -> Dict (Obj2 k a b)
```

Then `f &&& g` could get `Obj2 k a b` from `isObj f` and `Obj2 k b c` from `isObj g`, and hence provide the needed `Obj3 k a` to `dup`. I don’t know if this trick would work, but I think I was something like this in a category theory library by Sjoerd Visscher.

src/Category.hs
```diff
@@ -0,0 +1,189 @@
...
+             => (a `k` c) -> (a `k` d) -> (a `k` (c `p` d))
+pattern f :& g <- (unfork2 -> (f,g)) where (:&) = (&&&)
+-- {-# complete (:&) #-}
+
+-- GHC error:
+-- 
+--   A type signature must be provided for a set of polymorphic pattern synonyms.
+--   In {-# complete :& #-}
+--
+-- Given the generality of (:&), is there any solution to this completeness issue?
```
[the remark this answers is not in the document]
conal: Thanks\! I think I can live with that solution.

src/F.hs
```diff
@@ -0,0 +1,36 @@
...
+  type Obj (F s) = Representable 
+  id = F id
+  F g . F f = F (g . f)
+
+instance Monoidal (F s) (:*:) where
+  F f *** F g = F (\ (a :*: b) -> f a :*: g b)
+
+instance Cartesian (F s) (:*:) where
+  exl = F (\ (a :*: _) -> a)
+  exr = F (\ (_ :*: b) -> b)
```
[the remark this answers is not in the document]
conal: It’s bothered me before as well, so I added them in other projects. Nice choice for our custom prelude as you suggest.

src/Category.hs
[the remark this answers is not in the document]
dwincort: Oh I see!  That's clever!

conal:
I’ve been trying to maintain a high degree of generality in the category classes, which I anticipate providing most of the vocabulary for all of the linear map representations (including the semantics). Specifically,

  - Poly-kinded objects, including `*` and `* -> *`. We need the latter for linear maps. Well, “need” is too strong a word here. In concat I used `*` objects, and added a constraint to map between these types and representable functors. It seems much simpler to use representable functors directly.
  - Per-category binary products and coproducts. Ordinary functions use `(:+)` (`Either`) for coproducts, while linear algebra uses `(:*)`.
  - Per-category n-ary products and coproducts. With `*` objects, products look like `r a` (for representable `r` and object `a`), while functor category products look like `r :.: a`. Ordinary functions use an index/value pair for coproducts, while linear algebra uses one-hot vectors. (N-ary products an coproducts are *uniform*, though only because I don’t know how to type nonuniform versions nicely in Haskell.)

My intention is to define the meaning of the shared vocabulary on linear functions and then use it to specify everything else by saying that various simple isomorphisms among the representations are homomorphic with respect to this shared vocabulary. Then I think we’ll have precise and complete answers to most of the questions we’ve started discussing (other than performance) and will discuss.

I’m feeling weary right now as well as anxious about letting this category classes design go on, because it’s holding up the specification of everything else.

At this point, I’m tempted to remove generality, specializing the classes as follows:

  - Only functor categories, i.e., objects are always `* -> *`.
  - Only *biproduct* categories, so binary products and coproducts are always `(:*:)`, and n-ary products and coproducts are always `(:.:)`. Drop `(+++)` (coproduct bifunctor) in favor of always using `(***)`.

These specializations will still accommodate our linear map representations, because they all have the same meaning.

I think we can re-generalize later if we figure out how to with good usability.

Any thoughts?

conal:
directlyBTW, a lovely and useful of poly-kinded categories is making constraint entailment into a category. I got a lot of mileage out of constraint entailment programming in compiling-to-categories, but I had to wrap it all up in a `*`-kinded category instead of naturally and directly using the `Constraint` kind. The main use I made was things like convincing GHC of that `Obj k (a :* b)` entails `Obj k a` and `Obj k b`. Without that logic, utility function signatures required enormous constraints that reflected the functions’ *implementations* rather than their purposes.

src/Category.hs
```diff
@@ -0,0 +1,189 @@
...
+import qualified Prelude as P
+import Prelude hiding (id,(.))
+import GHC.Types (Constraint)
+import qualified Control.Arrow as A
+import Data.Functor.Rep
+
+import Misc
+
+class    Yes1 a
+instance Yes1 a
```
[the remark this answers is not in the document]
bolt12: Why not `()`? Isn't it simpler and shorter?

src/Category.hs
```diff
@@ -0,0 +1,189 @@
...
+-- {-# complete (:|) #-}  -- See (:&) above
+
+
+-- -- N-ary (representable) counterparts
+
+class (Category k, Representable r) => MonoidalR k r where
+  cross :: r (a `k` b) -> (r a `k` r b)
+
+-- TODO: maybe "N" (for n-ary) instead of "R" (for representable) in these
+-- class names. On the other hand, people will probably assume n is a number.
```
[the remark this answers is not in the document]
bolt12: I like `R,` also!

bolt12:
I think there's no harm in going for a less general solutil

src/Category.hs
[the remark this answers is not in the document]
dwincort: The problem is that `()` is kind `Constraint`, but we need a type of kind `u -> Constraint`.

An alternative is to change the definition in the `Category` class to
```haskell
type Obj k a :: Constraint
```
but this feels a little less clean.  Now, instead of being able to write
```haskell
type Obj (F s) = Representable
```
we have to write
```haskell
type Obj (F s) x = Representable x
```
Also, the `C2` through `C6` types used below will no longer work

src/Category.hs
```diff
@@ -0,0 +1,189 @@
...
+--   λ> :k Monoidal
+--   Monoidal :: (u -> u -> *) -> (u -> u -> u) -> Constraint
+--   λ> :k MonoidalR
+--   MonoidalR :: (* -> * -> *) -> (* -> *) -> Constraint
+--
+-- For L, we have
+-- 
+--   cross :: (...) => c (L a b s) -> L (c :.: a) (c :.: b) s
+--
+-- What generalization are we after here?
```
dwincort, line 123: This is an interesting question, and despite some fiddling, I couldn't come up with a particularly satisfying answer.  The root problem as I see it is that type functions in Haskell are just not quite where we need them to be.  If GHC supported [unsaturated type families](https://github.com/ghc-proposals/ghc-proposals/pull/242), we'd be fine, but until then, we can only make this properly kind-polymorphic with some annoying hacks.

For instance, consider the following class definitions:
```haskell
-- I'm not sure if we need/want to keep the Representable superclass constraint
class (Category k, Representable r) => MonoidalR k r p | k r -> p where
  cross :: r (a `k` b) -> (p r a `k` p r b)

class MonoidalR k r p => CartesianR k r p where
  exs :: r (p r a `k` a)
  dups :: a `k` p r a

fork :: (CartesianR k r p, Obj2 k a c) => r (a `k` c) -> (a `k` p r c)
fork fs = cross fs . dups

class (Category k, Representable r) => ComonoidalR k r p | k r -> p where
  plus :: r (a `k` b) -> (p r a `k` p r b)

class (Representable r, ComonoidalR k r p) => CocartesianR k r p where
  ins :: r (a `k` p r a)
  jams :: p r a `k` a
```
This actually has a really nice feature, which is that, like `Monoidal` and `Comonoidal` above, now these two classes directly mirror each other.  But what is `p`?

For `MonoidalR L`, we'll let `p` be `:.:`, and we're good to go, but the choice of `p` for `ComonoidalR L` is not quite as clean.  Specifically, I think we'll need to define:
```haskell
newtype RepPair r x = RepPair { unRepPair :: (Rep r :* x)}
```

Additionally, we'll need newtypes for both the `MonoidalR (->)` and `ComonoidalR (->)` instances.  The code I laid out looks like:
```haskell
newtype Ap f a = Ap { unAp :: f a }
  deriving (Functor, Generic, Generic1)

instance Distributive f => Distributive (Ap f) where
  distribute = Ap . distribute . fmap unAp

instance Representable f => Representable (Ap f) where
  type Rep (Ap f) = Rep f
  tabulate = Ap . tabulate
  index = index . unAp

newtype RepPair r x = RepPair { unRepPair :: (Rep r :* x)}

instance Representable p => MonoidalR (->) p Ap where
  cross rab (Ap ra) = Ap $ liftR2 ($) rab ra

instance Representable p => CartesianR (->) p Ap where
  exs = tabulate (flip index)
  dups = pureRep

instance Representable p => ComonoidalR (->) p RepPair where
  plus fs (RepPair (i,a)) = RepPair (i, (fs `index` i) a)

instance Representable p => CocartesianR (->) p RepPair where
  ins = tabulate $ (RepPair .) . (,)  -- = tabulate (\ i a -> RepPair (i, a))
  jams = snd . unRepPair
```

src/Category.hs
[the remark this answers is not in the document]
bolt12: I see, thanks!

src/Category.hs
dwincort: Just for kicks, I added instances for my modified versions of `MonoidalR` and similar for `F`:
```haskell
instance Representable r => MonoidalR (F s) r (:.:) where
  -- cross :: r (F s a b) -> F s (r :.: a) (r :.: b)
  cross rf = F $ Comp1 . liftR2 unF rf . unComp1

instance Representable r => CartesianR (F s) r (:.:) where
  -- exs :: r (F s (r :.: a) a)
  exs = tabulate $ \repr -> F $ flip index repr . unComp1
  -- dups :: F s a (r :.: a)
  dups = F $ Comp1 . pureRep

newtype RepPair1 r f s = RepPair1 { unRepPair1 :: (Rep r * f s)}

mapRepPair1 :: (Rep r -> f s -> g t) -> RepPair1 r f s -> RepPair1 r g t
mapRepPair1 f (RepPair1 (repr, fs)) = RepPair1 (repr, f repr fs)

instance Representable r => ComonoidalR (F s) r RepPair1 where
  -- plus :: r (F s a b) -> F s (RepPair1 r a) (RepPair1 r b)
  plus rf = F $ mapRepPair1 $ unF . index rf

instance Representable r => CocartesianR (F s) r RepPair1 where
  -- ins :: r (F s a (RepPair1 r a))
  ins = tabulate $ F . (RepPair1 .) . (,)
  -- jams :: F s (RepPair1 r a) a
  jams = F $ snd . unRepPair1
```

conal: @dwincort: Wow. Thanks very much for working out and sharing all this detail\! I don’t think I could any better. Hopefully better future possibilities with improved GHC, as you mention. (Thanks also for the encouraging pointer to unsaturated type application work.)

conal:
I’ve committed a bunch of work toward a common vocabulary (in `Category`) for all linear map representations. At this point only `F` uses this vocabulary, so it may need some tweaking. In particular, the classes have wired in that the representable functors h used in `join` and `fork` are constrained by `(Representable h, Foldable h, Eq (Rep h)`. If we want something different for `L` (as currently) and/or other representations, we can add another associated constraint. I didn’t generalize in this way yet, because I’m anticipating a fresh re-examination of `L` will occur during the vocabulary conversion.

The method `mu` introduced in `F` is the semantic function and associated correctness specification *for all representations*. For `L`,

``` haskell
instance Linear s L (:*:) where
  mu (Scale s)  = scale s
  mu (f :|# g)  = mu f ||| mu g
  mu (f :&# g)  = mu f &&& mu g
  mu (JoinL fs) = join (mu <$> fs)
  mu (ForkL fs) = fork (mu <$> fs)
```

I expect that we’ll define `scale = Scale`, `(|||) = (:|#)`, …, `fork = ForkL`, from which the corresponding required homomorphism properties follow immediately.

In addition to `L`, I’m hoping we’ll define a few more representations, including the following:

  - *Non-inductive* matrices in row major and column major form, i.e., `g :.: f` and `f :.: g` to represent `F s f g`, with `g` and `f` being representable functors.
  - Same but represented as `Vec (Card (Rep g)) :.: Vec (Card (Rep f))` and `Vec (Card (Rep f)) :.: Vec (Card (Rep g))`, where `Card` is a cardinality type family and `Vec` is an efficiently implemented type of sized vectors like the one in vector-sized. Later, I want to switch to much more efficient, non-CPU implementations.

conal:
There’s another wonderful tool for making linear map representations, namely the *opposite* of any other representation. I mean “opposite” in the sense of the opposite category construction, representing `a :-* b` by `b :-* a`, swapping joins and forks, and flipping compositions. For linear maps this construction also and maps scalings to themselves.

If we start with a matrix representation, transposition emerges, as we’ve seen in other conversations. If instead we start with the semantics itself (`F`), we get the linear map representation that leads to simple, efficient, purely functional (and thus parallel-friendly) exact gradient computation and thus deep learning. In terms of linear algebra, `b :-* a` really represents `DualV b :-* DualV a`, where `DualV z` is “linear functionals”, i.e., linear maps from `z` to its scalar semirings, which happen to be isomorphic to `z` itself.

This whole idea generalizes beyond efficient gradient computation (“functional backpropagation”) to “reverse-mode automatic differentiation” by the construction that Cayley invented for group theory, which generalizes to the continuation-passing transformation, and much more to the Yoneda embedding. (See [*The simple essence of automatic differentiation*](http://conal.net/papers/essence-of-ad), sections 12 & 13.)

conal:
A flaw of the “semantic representation” (`F`) is that linearity is implicit and not checked by the (Haskell) compiler. It would be lovely to rebuild all of this stuff in Agda or Idris so that linearity is guaranteed by the compiler.

conal:
Conjecture: the opposite category construction also turns the row-major matrix implementations into column-major and vice versa.

bolt12:
I tried to see how it would go if I tried to give a `Category` instance to `L` and swapped the type arguments so it would fit the kind of the type class, however I stumbled upon this error which I'm not sure how to overcome:

```haskell
instance C.Category (L s) where
  type Obj (L s) = C
  id = idL
  (.) = (.@)
```

```haskell
The type synonym ‘V’ should have 1 argument, but has been given none
```

`V` is a type synonym and cannot be partially applied... is there a way to go around this?

conal:
> `V` is a type synonym and cannot be partially applied… is there a way to go around this?

Yes. Before we had constraint kinds, we had to define new classes and a universal instance:

``` haskell
type V' f = ((Representable f, Foldable f, Eq (Rep f), ToScalar f, ToRowMajor f)
             :: Constraint)

class    V' f => V f
instance V' f => V f
```

The `V'` definition here is just a convenience to avoid replicating constraint lists.

bolt12:
Oh right, this helps! But now what about the composition? I needs proper constraints on `L` as well, should the `(.)` be changed in order to add this too?

dwincort:
> Oh right, this helps! But now what about the composition? I needs proper constraints on `L` as well, should the `(.)` be changed in order to add this too?

An alternative to needing constraints on `L` is that `L` should have all of the information necessary _within_ it.  In other words, `L` should be defined such that it is guaranteed that `V f` and `V g` are satisfied.  A close alternative to a restriction like this is outlined in [an earlier comment by @conal](https://github.com/conal/linalg/pull/28#discussion_r464749511).

For the problem at hand, I wonder if it's possible to rewrite `.@` for `L` so that we can pattern match in such a way as to make the `V3 f g h` requirement in the type unnecessary.

conal:
> Oh right, this helps\! But now what about the composition? I needs proper constraints on `L` as well, should the `(.)` be changed in order to add this too?

Maybe so. [In concat](https://github.com/conal/concat/blob/master/classes/src/ConCat/Category.hs) I made these constraints explicit on every operation (method and utility function). I’ve seen other similar designs that had them only for operations without morphism arguments, and I’m curious to learn how it goes. So I guess there’s a choice here of whether to proceed as in concat or try something new.

bolt12:
Unfortunately it does not compile on my end:

```haskell
• Could not deduce (Eq (Rep h2)) arising from a use of ‘:&#’
```

```haskell
• Could not deduce (Foldable h1) arising from a use of ‘:|#’
```

```haskell
• Could not deduce (Eq (Rep h2)) arising from a use of ‘ForkL’
```


```haskell
• Could not deduce (Foldable h2) arising from a use of ‘JoinL’
```

src/F.hs
```diff
@@ -0,0 +1,93 @@
...
+  scale s = F (fmap (s *))   -- F (\ (Par1 s') -> Par1 (s * s'))
+
+class Linear l where
+  -- | Semantic function for all linear map representations. Correctness of
+  -- every operation on every representation is specified by requiring that mu
+  -- is homomorphic for (distributes over) that operation. For instance, mu must
+  -- be a functor (Category homomorphism).
+  mu  :: l s a b -> F s a b
+  -- | Inverse of mu
+  mu' :: F s a b -> l s a b
```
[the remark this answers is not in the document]
conal: > Should this be in the `Category` module?

I don’t think it can be, since `F` depends on `Category`, and putting `Linear` in `Category` would make it depend on on `F`.

I also don’t think it should be, since `Category` is more general than linear maps. Maybe I should move `Scalable` from `Category` to `F`, however.

> Also probably we might want to add constraints to the context of `mu` and `mu'` since the categorical vocabulary has some.

I think so, too, unless we decide to extract `Obj k a` and `Obj k b` from every ``a `k` b`` [as in that earlier conversation Dan mentioned](https://github.com/conal/linalg/pull/28#discussion_r464749511).

conal:
I just committed an experiment along the lines of [the earlier discussion of extracting constraint proofs from morphisms](https://github.com/conal/linalg/pull/28#discussion_r464749511):

``` haskell
-- Prove V/Obj constraints from morphisms
isV :: L a b s -> Dict (V2 a b)
isV (Scale _) = Dict
isV (_ :|# _) = Dict
isV (_ :&# _) = Dict
isV (JoinL _) = Dict
isV (ForkL _) = Dict

infixr 9 @.@
(@.@) :: Semiring s => L b c s -> L a b s -> L a c s
b @.@ a | (Dict, Dict) <- (isV a, isV b) = b .@ a

(.@) :: (V3 a b c, Semiring s) => L b c s -> L a b s -> L a c s
...
```

conal:
Or

``` haskell
b@(isV -> Dict) @.@ a@(isV -> Dict) = b .@ a
```

Better yet,

``` haskell
pattern IsV <- (isV -> Dict)
{-# COMPLETE IsV :: L #-}

pattern IsV :: () => V2 a b => L a b s
b@IsV @.@ a@IsV = b .@ a
```

I got that `IsV` type from GHCi. @dwincort: do you understand the purpose of the (necessary) “`() =>`”?

conal:
And then generalize to an `isObj` method on Category and corresponding `IsObj` constraint.

I don’t know if we want to go down this route, since we’d still need explicit constraints for operations without morphism parameters (`id`, `exl`, `exr`, etc).

dwincort:
> Better yet,
> 
> ```haskell
> pattern IsV <- (isV -> Dict)
> {-# COMPLETE IsV :: L #-}
> 
> pattern IsV :: () => V2 a b => L a b s
> b@IsV @.@ a@IsV = b .@ a
> ```

Yes!  That pattern is what I was thinking, but I was thinking of using it directly in `.@`, as in:
```haskell
(.@) :: (Semiring s) => L g h s -> L f g s -> L f h s
Scale a   .@ Scale b   = Scale (a * b)             -- Scale denotation
(p :&# q) .@ m@IsV     = (p .@ m) :&# (q .@ m)
m@IsV     .@ (p :|# q) = (m .@ p) :|# (m .@ q)
(r :|# s) .@ (p :&# q) = (r .@ p) + (s .@ q)       -- biproduct law
ForkL ms' .@ m@IsV     = ForkL (fmap (.@ m) ms')
m'@IsV    .@ JoinL ms  = JoinL (fmap (m' .@) ms)
JoinL ms' .@ ForkL ms  = sum (ms' .^ ms)           -- biproduct law
```

> I got that `IsV` type from GHCi. @dwincort: do you understand the purpose of the (necessary) “`() =>`”?

Pattern type signatures are a little weird.  The set of constraints before the first `=>` are the ones that are _required_ for the pattern to apply.  The set of constraints before the second `=>` are the ones that the pattern _proves when it applies_.  If only one set of constraints is given, it is assumed that it's the _first_ set — in other words, you can easily declare requirements without needing to explicitly declare any proofs.  However, if you have no requirements but you do have proofs, then you need to have both clauses.  This is the case you have here: no requirements (hence the `() =>`) but still something proved (`V2 a b`).

conal:
> Yes\! That pattern is what I was thinking, but I was thinking of using it directly in `.@`, as in: \[…\]

Oh, nicer\!

> Pattern type signatures are a little weird. \[…\]

Got it. Thanks.

conal:
> Yes\! That pattern is what I was thinking, but I was thinking of using it directly in `.@`, as in: \[…\]

I guess a drawback is that `isV` gets applied recursively (I think) instead of just at the top. Or can GHC somehow eliminate these run-time tests altogether, given that they’re non-recursive and so always evaluate to `Dict` (and never to bottom)? I added `{-# OPTIONS_GHC -O2 -ddump-simpl #-}` to `LinAlg` and recompiled. It looks like `isV` is still doing its work.

dwincort:
> It looks like `isV` is still doing its work.

Yes, it seems you're right.  It really feels like there should be a way to convince GHC to skip the runtime tests, but I don't know how.  Well, the `@.@` helper functions works well, or we can use `IsV` directly in the definition of `(.)` and avoid having the extra operator.

bolt12:
So I gave implementing the `Category` instance for `L` a chance and for the most part it went very smoothly! Just a few barriers regarding constraints.

Only changes I did to the code were:

- Change the order the type parameters of `L`
- Define `(***)` with the help of `IsV` and reduce the constraints:
```haskell
(***) :: Additive s
      => L s a c -> L s b d -> L s (a :*: b) (c :*: d)
f@IsV *** g@IsV = (f :|# zero) :&# (zero :|# g)
```

Then the code for the instances:

```haskell
class    V f => V' f
instance V f => V' f

instance Semiring s => C.Category (L s) where
  type Obj (L s) = V'
  id = idL
  (.) = (@.@)

instance Semiring s => C.Monoidal (L s) (:*:) where
  (***) = (***)

instance Semiring s => C.Cartesian (L s) (:*:) where
  exl = exl
  exr = exr
  dup = idL :&# idL

instance Semiring s => C.Comonoidal (L s) (:*:) where
  (+++) = (***)

instance Semiring s => C.Cocartesian (L s) (:*:) where
  inl = inl
  inr = inr
  jam = idL :|# idL

instance Semiring s => C.Biproduct (L s) (:*:)

instance Semiring s => C.MonoidalR (L s) where
  cross = undefined -- cross
  -- • Could not deduce (Representable a) arising from a
  -- use of ‘cross’

instance Semiring s => C.CartesianR (L s) where
  exs = undefined -- exs
  dups = undefined -- ForkL idL
  -- See above. It needs some kind of constraints also

instance Semiring s => C.BiproductR (L s) where
  ins = undefined -- ins
  jams = undefined -- JoinL idL
  -- See above. It needs some kind of constraints also

instance Semiring s => C.Scalable s L (:*:) where
  scale = Scale

instance F.Linear L where
  mu (Scale s) = F.F (\_ -> Par1 s)
  mu (f :|# g) = undefined -- (F.mu f) C.||| (F.mu g) -- Needs (Additive s)
  mu (f :&# g) = F.mu f C.&&& F.mu g
  mu (JoinL ms) = undefined -- C.join (fmap F.mu ms) -- Needs (Additive s)
  mu (ForkL ms) = undefined -- C.fork (fmap F.mu ms) -- Needs (Additive s)

  mu' (F.F f) = undefined -- rowMajToL . distribute $ f <$> lToRowMaj idL
  -- Needs a bunch of constraints (Representable, Functor, Semiring)
```

conal:
> So I gave implementing the `Category` instance for `L` a chance and for the most part it went very smoothly\! Just a few barriers regarding constraints.

I’ll finish adding constraints to the operations in `Category` and let you know.

dwincort:
Huh, `cross` is a particularly annoying case.  It really feels like I should be able to write:
```haskell
cross :: (V c, Additive s) => c (L a b s) -> L (c :.: a) (c :.: b) s
cross x | Dict <- foldMap isV x = JoinL $ fmap ForkL $ diagRep zero x
```
but it doesn't work.  Specifically, if `c` is something like `Const ()`, then there really is no `L a b s` to extract the `V2 a b` from.  (Also, we're given `Representable c` but not `Foldable c`.)

conal:
Oh, hey. Good point, @dwincort\!

src/F.hs
```diff
@@ -0,0 +1,93 @@
...
+
+instance Additive s => BiproductR (F s) where
+  ins :: (C2 Representable r a, Eq (Rep r)) => r (F s a (r :.: a))
+  ins = tabulate (F . oneHot)
+        -- tabulate $ \ i -> F (oneHot i)
+        -- tabulate $ \ i -> F (\ a -> oneHot i a)
+  jams :: (C2 Representable r a, Foldable r) => F s (r :.: a) a
+  jams = F (\ (Comp1 as) -> foldr (+^) zeroV as)
+
+-- TODO: can we define ins without Eq (Rep r)?
```
dwincort, line 69: Didn't you solve this in your "Compiling to Categories" paper by introducing the `Diagonal` type class?  I can't find a full definition of said type class in your paper, but I imagine it would be something like:
```haskell
class Representable h => Diagonalizable h where
  diag :: a -> h a -> h (h a)
```
We can make this a little nicer by including a default definition when we know `Eq (Rep r)` (i.e., `diagRep`), but we don't need that constraint to define `diag` for many forms.  For instance:
```haskell
instance (Diagonalizable f, Diagonalizable g) => Diagonalizable (f :*: g) where
  diag dflt (f :*: g) =
    ((:*: pureRep dflt) <$> diag dflt f) :*:
    ((pureRep dflt :*:) <$> diag dflt g)

instance (Diagonalizable f, Diagonalizable g) => Diagonalizable (f :.: g) where
  diag dflt (Comp1 fg) = fmap Comp1 $ Comp1 $ fmap distribute $ diag (pureRep $ pureRep dflt) $ diag dflt <$> fg
```
(I think those are right.)

conal:
Yesterday I added constraints to all of the methods in `Category` (as we discussed in our video meeting) and immediately re-encountered a sticky issue.

Consider, for instance, the following updated `Category`, `Monoidal`, and `Cartesian` classes:

``` haskell
class Category (k :: u -> u -> *) where
  type Obj k :: u -> Constraint
  infixr 9 .
  id :: Obj k a => a `k` a
  (.) :: Obj3 k a b c => (b `k` c) -> (a `k` b) -> (a `k` c)

class Category k => Monoidal k p | k -> p where
  infixr 3 ***
  (***) :: Obj4 k a b c d => (a `k` c) -> (b `k` d) -> ((a `p` b) `k` (c `p` d))

class Monoidal k p => Cartesian k p where
  exl :: Obj2 k a b => (a `p` b) `k` a
  exr :: Obj2 k a b => (a `p` b) `k` b
  dup :: Obj  k a   => a `k` (a `p` a)
```

Define binary fork in terms of this vocabulary:

``` haskell
infixr 3 &&&
(&&&) :: (Cartesian k p, Obj3 k a c d)
      => (a `k` c) -> (a `k` d) -> (a `k` (c `p` d))
f &&& g = (f *** g) . dup
```

GHC objects:

    Could not deduce: (Obj k (p a a), Obj k (p c d))
      arising from a use of ‘.’
    from the context: (Cartesian k p, Obj3 k a c d)

The issue here is that `(.)` now requires the result of `dup`, ``a `p` a``, to be a valid `k`-object, which is not provable from `Obj3 k a b c`.

This issue was the first hard challenge in my compiling-to-categories implementation, leading to the choice of its name “concat” for “constrained categories”. My solution was to design, implement, and use a sort of DSL for constraint entailment proofs in a form GHC can use for type-checking. Delightfully, constraint entailment forms a cartesian category, so I used the same categorical vocabulary to express the typing proofs that GHC needs to type-check the implementation of that vocabulary. I learned a lot in the process and haven’t written it up yet other than a brief mention in [*Compiling to categories*](http://conal.net/papers/compiling-to-categories), section 7.8. I’ve since thought there may now be better solutions:

  - A type-checker plugin that can automate these proofs in a systematic way, i.e., more generally than this one logic or even the language/classes in which the logic is defined.
  - Take advantage of recent GHC improvements in expressiveness of constraints, particularly universal quantification and implication constraints.

I would love to have help exploring and weighing design options. I expect that the discussion will lead us on a digression from linear algebra, though I hope illuminating to all involved.

Reactions, please.

dwincort:
> * Take advantage of recent GHC improvements in expressiveness of constraints, particularly universal quantification and implication constraints.

I didn't know about those, but after [reading the docs](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#extension-QuantifiedConstraints), they do seem like they'd do the trick.  Unfortunately, the obvious way to use them fails.  That is, if you change the definition of the `Monoidal` class to:
```haskell
class (forall a b. Obj2 k a b => Obj k (a `p` b), Category k) => Monoidal k p | k -> p where
```
then you get an error:
```
    • Quantified predicate must have a class or type variable head:
        forall (a :: u) (b :: u). Obj2 k a b => Obj k (p a b)
    • In the quantified constraint ‘forall (a :: u) (b :: u).
                                    Obj2 k a b =>
                                    Obj k (p a b)’
      In the context: (forall (a :: u) (b :: u).
                       Obj2 k a b =>
                       Obj k (p a b),
                       Category k)
      While checking the super-classes of class ‘Monoidal’
      In the class declaration for ‘Monoidal’
```
Apparently, type synonym constraints are not allowed as the predicate head for quantified predicates, which honestly seems like a weird oversight to me (although maybe there's a good reason that I'd discover if I read the paper).  However, we can get around this.  Instead of making `Obj` an associated type synonym, we can make it part of the class head and use functional dependencies to fix its type.  I'm not sure if this has down-the-line repercussions, but the following code type-checks:
```haskell
class Category (k :: u -> u -> *) (obj :: u -> Constraint) | k -> obj where
  infixr 9 .
  id :: obj a => a `k` a
  (.) :: C3 obj a b c => (b `k` c) -> (a `k` b) -> (a `k` c)

class (forall a b. C2 obj a b => obj (a `p` b), Category k obj) => Monoidal k obj p | k -> p obj where
  infixr 3 ***
  (***) :: C4 obj a b c d => (a `k` c) -> (b `k` d) -> ((a `p` b) `k` (c `p` d))

class Monoidal k obj p => Cartesian k obj p where
  exl :: C2 obj a b => (a `p` b) `k` a
  exr :: C2 obj a b => (a `p` b) `k` b
  dup ::    obj a   => a `k` (a `p` a)

infixr 3 &&&
(&&&) :: (Cartesian k obj p, C3 obj a c d)
      => (a `k` c) -> (a `k` d) -> (a `k` (c `p` d))
f &&& g = (f *** g) . dup
```

bolt12:
From my experiments, only the `Linear` and `R` classes struggled with constraints. If we want we can just fix the signatures of those classes' methods and try to come up with a better solution regarding overall constraints later.

I don't really like having the `obj` constraint as a part of the class head but I'm okay with it if we see that that is the best solution!

dwincort:
Hmm, if we turn on `UndecidableInstances` and `UndecidableSuperClasses`, then we perhaps have a less ugly solution.  The problem I ran into with `QuantifiedConstraints` was that I couldn't use a type synonym as the head of the predicate.  So, why not turn the synonym into a class?
```haskell
class Obj' k a => Obj k a
instance Obj' k a => Obj k a

class Category (k :: u -> u -> *) where
  type Obj' k :: u -> Constraint
  ...

class (forall a b. Obj2 k a b => Obj k (a `p` b), Category k) => Monoidal k p | k -> p where
  ...
```

---

> If we want we can just fix the signatures of those classes' methods

@bolt12 Isn't this what we're trying to do?  When we add constraints to those functions, there are cascading effects that affect, e.g., the definition of `(&&&)`.

src/Category.hs
```diff
@@ -0,0 +1,192 @@
...
+  infixr 3 ***
+  (***) :: (a `k` c) -> (b `k` d) -> ((a `p` b) `k` (c `p` d))
+
+-- The functional dependency requires p to be uniquely determined by k. Might
+-- help type inference. Necessitates a "Comonoidal" class with "(+++)", which is
+-- perhaps better than giving two Monoidal instances for a single category (eg
+-- for (->)).
+
+-- TODO: make p an associated type, and see how the class and instance
+-- definitions look in comparison.
```
dwincort, line 48: From what I can tell, if we use `QuantifiedConstraints` with `p`, then we can't turn it into an associated type.  I'm not sure that's so bad, but it's worth noting.

bolt12:
@dwincort Yes, but what I'm saying is that I don't think it is necessary to constrain `(.)`, `Monoidal` and `Cartesian` when only the `R` versions of those classes require tweaks. It might be a little too ad-hoc though so I understand if this is not the preferable solution.

src/F.hs
conal: > > `TODO: can we define ins without Eq (Rep r)?`
> 
> Didn’t you solve this in your “Compiling to Categories” paper by introducing the `Diagonal` type class? I can’t find a full definition of said type class in your paper, but I imagine it would be something like:
> 
> ``` haskell
> class Representable h => Diagonalizable h where
>   diag :: a -> h a -> h (h a)
>   default diag :: Eq (Rep h) => a -> h a -> h (h a)
>   diag dflt as =
>     tabulate (\ i -> tabulate (\ j -> if i == j then as `index` i else dflt))
> ```

Yes, indeed, and thanks for this reminder. There are two implementations in [`ConCat.Free.Diagonal`](https://github.com/conal/concat/blob/master/examples/src/ConCat/Free/Diagonal.hs), with one as a new `Diagonal` class and the other via existing classes, roughly corresponding to your definition above and to the one depending on `Eq (Rep h)` but with more leeway for efficient implementation.

conal: Hm. If we have to constrain `h`, then relying on general, well-established algebraic abstractions like `Category`, `Representable` and `Eq` has some appeal to me, including less invention of concept and implementation.

Other perspectives?

conal:
> Apparently, type synonym constraints are not allowed as the predicate head for quantified predicates, which honestly seems like a weird oversight to me (although maybe there’s a good reason that I’d discover if I read the paper). However, we can get around this. Instead of making `Obj` an associated type synonym, we can make it part of the class head and use functional dependencies to fix its type.

Hm\! I don’t think I ever considered shifting the associated `Obj` constraint into the class parameters. Neat\! I also didn’t know about that “class or type variable head” restriction on quantified predicates.

> Hmm, if we turn on `UndecidableInstances` and `UndecidableSuperClasses`, then we perhaps have a less ugly solution. The problem I ran into with `QuantifiedConstraints` was that I couldn’t use a type synonym as the head of the predicate. So, why not turn the synonym into a class? \[…\]

I like this version\!

conal:
I’m delighted with this quantified constraints formulation\! See commit https://github.com/conal/linalg/pull/28/commits/84cea551b0bef92b2df0ae64bf2e2e317acdb645. So much simpler than with constraint entailment programming.

Thanks, @dwincort for scouting out this idea. :D

src/F.hs
dwincort: My only counter thought is that I find it interesting that I can define `Diagonalizable` instances for the standard generic types without ever needing `Eq (Rep h)`.  I guess this is a little disingenuous — we _are_ relying on the fact that `()` and `Void` have `Eq` instances and that they lift naturally to sums and products — but still, it feels frustrating that we have to have the `Eq (Rep h)` constraint at all.

Maybe making a `Diagonal` type class is a step too far.  Maybe all we need is good naming:
```haskell
type Diagonal h = (Representable h, Eq (Rep h))
type CountablyRepresentable h = (Representable h, HasFin (Rep h))
```
Leaving out type classes makes it clear that these are not whole new concepts, but giving the constraints names also lends some weight to the idea behind them.

Perhaps another interesting line of inquiry might be: If `Diagonalizable` is an invented concept (and I now think I agree that it is), then is there a more specialized form of `Representable` that _isn't_ merely an invention?

conal: > My only counter thought is that I find it interesting that I can define `Diagonalizable` instances for the standard generic types without ever needing `Eq (Rep h)`.

Ah; I like this point.

> I guess this is a little disingenuous — we *are* relying on the fact that `()` and `Void` have `Eq` instances and that they lift naturally to sums and products — but still, it feels frustrating that we have to have the `Eq (Rep h)` constraint at all.

Hm\! Interesting angle about `()` and `Void`. Funny to think of `Void` having an `Eq` instance, but it certainly does. :) I share your distaste for `Eq (Rep h)`. It’s unnecessary and motivated solely by implementation, and it commits us to inefficiency.

> Maybe making a `Diagonal` type class is a step too far. Maybe all we need is good naming:
> 
> ``` haskell
> type Diagonal h = (Representable h, Eq (Rep h))
> type CountablyRepresentable h = (Representable h, HasFin (Rep h))
> ```
> 
> Leaving out type classes makes it clear that these are not whole new concepts, but giving the constraints names also lends some weight to the idea behind them.

Sure. I like how this choice insulates the rest of the implementation from `Eq (Rep h)`.

Side comment: I always use and understand “countable” to mean finite or countably infinite, especially the latter. “`FinitelyRepresentable`” seems more accurate, but that term already has another meaning.

> Perhaps another interesting line of inquiry might be: If `Diagonalizable` is an invented concept (and I now think I agree that it is), then is there a more specialized form of `Representable` that *isn’t* merely an invention?

***This question exactly\!***

conal:
I added constraint entailment just because it's pretty and it shows off poly-kinded categories.

src/Category.hs
dwincort: When I try it, I get:
```
    • Illegal type synonym family application ‘P k’ in instance:
        Obj k (P k a b)
    • In the quantified constraint ‘forall (a :: u) (b :: u).
                                    Obj2 k a b =>
                                    Obj k (P k a b)’
      In the context: (forall (a :: u) (b :: u).
                       Obj2 k a b =>
                       Obj k (P k a b),
                       Category k)
      While checking the super-classes of class ‘Monoidal’
      In the class declaration for ‘Monoidal’
   |
42 | class (forall a b. Obj2 k a b => Obj k (P k a b), Category k) => Monoidal k where
```
I'm not sure how to get around that error.

conal:
I’ve gotten to a stable place, so maybe now is a good time to merge the *specification* branch into *master*. Then we can recreate the functionality of `LinAlg` in `InductiveMatrix`, discontinue `LinAlg`, and add other linear map and other operations. Also pick through discussions and improvements from other issues, reviews, and pull requests.

I’m happy to continue filling in `InductiveMatrix`, but it might be more constructive and fun for others to do so. The specification is always that the semantic function is homomorphic over the implemented vocabulary.

src/InductiveMatrix.hs
```diff
@@ -0,0 +1,58 @@
...
+-- | Representation and its denotation
+-------------------------------------------------------------------------------
+
+infixr 3 :&#
+infixr 2 :|#
+
+type V = Representable
+
+-- | Compositional linear map representation.
+data L :: * -> (* -> *) -> (* -> *) -> * where
```
dwincort, line 37: While you're renaming and rewriting, perhaps make all the data types have the same name (distinguishable via which one gets imported or by qualified imports).  `L` for "linear map" is pretty good, so perhaps nothing here needs to change, but then that means we should change `F` in `LinearFunction` to be `L` also.

conal: Done. Thanks for the reminder!

PR #30, opened 2020-07-29
Fewer additives

conal:
I removed several unnecessary `Additive` constraints and downgraded one `Semiring` to `Additive`. I think the remaining `Additive` and `Semiring` constraints are all necessary. Thanks for pointing out fishy constraints, and please keep it up.

### DIFF
--- a/src/LinAlg.hs
+++ b/src/LinAlg.hs
@@ -63,7 +63,7 @@ data L :: (* -> *) -> (* -> *) -> (* -> *) where
 
 unjoin2 :: V3 f g h => L (f :*: g) h s -> L f h s :* L g h s
 unjoin2 (p :|# q) = (p,q)
-unjoin2 ((unjoin2 -> (p,q)) :&# (unjoin2 -> (r,s))) = (p :&# r, q :&# s)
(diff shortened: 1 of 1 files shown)


src/LinAlg.hs
```diff
@@ -61,9 +61,9 @@ data L :: (* -> *) -> (* -> *) -> (* -> *) where
   JoinL :: V3 f g h => h (L f g s) -> L (h :.: f) g s
   ForkL :: V3 f g h => h (L f g s) -> L f (h :.: g) s
 
-unjoin2 :: (V3 f g h, Additive s) => L (f :*: g) h s -> L f h s :* L g h s
+unjoin2 :: V3 f g h => L (f :*: g) h s -> L f h s :* L g h s
 unjoin2 (p :|# q) = (p,q)
-unjoin2 ((unjoin2 -> (p,q)) :&# (unjoin2 -> (r,s))) = (p :& r, q :& s)
+unjoin2 ((unjoin2 -> (p,q)) :&# (unjoin2 -> (r,s))) = (p :&# r, q :&# s)
```
dwincort, line 66: I'm perfectly happy with this change, but I will note that now that the type of the `:&` pattern doesn't require `Additive`, this change is unnecessary.

conal: Thanks! Fixed now. Next time I'll read my own diffs more carefully.

PR #32, opened 2020-07-30
Failable patterns redux

dwincort:
This PR subsumes https://github.com/conal/linalg/pull/27.

I realized I could use a singletons-like trick to get the type information I needed to do failable patterns without resorting to crazy GHC hacks.  The result is shown in this PR.

Thus, 2 of the 4 downsides of the linked PR have been addressed: `Zero` is no longer a constructor, and the implementation is no longer ugly.  The complete pragmas are still insufficient (https://gitlab.haskell.org/ghc/ghc/-/issues/16892), but that's not a big deal — these new patterns can be used in conjunction with the more restricted patterns.  Also, there's still a weird type-checker slow-down that's introduced by the complete pragma, but it's a few seconds as opposed to a few minutes, so perhaps it's manageable.


src/LinAlg.hs
```diff
@@ -19,15 +18,13 @@ type (:*)  = (,)
...
-            :: Constraint)
+type V' f = (Representable f, Foldable f, Eq (Rep f), ToScalar f, ToRowMajor f)
+type V  f = (V' f, HasShape f)
 type V2 f g = (V f, V g)
 type V3 f g h = (V2 f g, V h)
+type V3' f g h = (V2 f g, V' h)
 type V4 f g h k = (V2 f g, V2 h k)
 
--- TODO: Why does V suddenly need ":: Constraint" after adding ToScalar f and
--- ToRowMajor f?
```
dwincort, line 29: I don't know why, and when i tried removing `:: Constraint`, it still works for me.

.gitignore
```diff
@@ -1 +1,3 @@
 linalg.cabal
+.stack*
+stack.yaml.lock
```
dwincort, line 3: Even if this PR doesn't get merged, we should update the .gitignore for these stack files.

src/LinAlg.hs
```diff
@@ -58,8 +55,8 @@ data L :: (* -> *) -> (* -> *) -> (* -> *) where
   Scale :: s -> L Par1 Par1 s
   (:|#) :: V3 f g h => L f h s -> L g h s -> L (f :*: g) h s
   (:&#) :: V3 f h k => L f h s -> L f k s -> L f (h :*: k) s
-  JoinL :: V3 f g h => h (L f g s) -> L (h :.: f) g s
-  ForkL :: V3 f g h => h (L f g s) -> L f (h :.: g) s
+  JoinL :: V3' f g h => h (L f g s) -> L (h :.: f) g s
+  ForkL :: V3' f g h => h (L f g s) -> L f (h :.: g) s
```
dwincort, line 59: I need to know the shape `f` and `g`, but I'm happy for `h` to be arbitrary.  Thus, I created a new constraint `V'` that is different from `V` in that `V` has a `HasShape` constraint while `V'` doesn't.  Furthermore, `V3'` is `V` of the first two arguments and `V'` of the third.

src/LinAlg.hs
conal: Really?! With which version of GHC?

dwincort: Honestly, I have no idea why it's working now.  I just played around with this a bunch and am thoroughly confused.  On `master`, I can't make the code compile without `:: Constraint`, but on this `failable-patterns-redux` branch, it works fine.  Perhaps it's something to do with the addition of `HasShape`?  I'm at a loss.

conal: Odd. Thanks for fiddling and letting me know.

.gitignore
conal: How do you all decide what patterns to put in a per-project .gitignore and what in your per-user global .gitignore? I have the two additional in the latter, along with many other patterns.

dwincort: Oh.  I don't have a per-user global .gitignore.  I guess it means I have a lot of duplication among my per-project ones, but I like the consistency of not using user-level customization.  In other words, if we put these two lines in the project's .gitignore, then none of us will be tempted to commit these files regardless of what our personal user-level .gitignore says.

bolt12: I usually use `summoner` to generate the Haskell project boilerplate for me and it uses a very complete .gitignore:

```
### Haskell
dist
dist-*
cabal-dev
*.o
*.hi
*.chi
*.chs.h
*.dyn_o
*.dyn_hi
*.prof
*.aux
*.hp
*.eventlog
.virtualenv
.hsenv
.hpc
.cabal-sandbox/
cabal.sandbox.config
cabal.config
cabal.project.local
.ghc.environment.*
.HTF/
# Stack
.stack-work/
stack.yaml.lock

### IDE/support
# Vim
[._]*.s[a-v][a-z]
[._]*.sw[a-p]
[._]s[a-v][a-z]
[._]sw[a-p]
*~
tags

# IntellijIDEA
.idea/
.ideaHaskellLib/
*.iml

# Atom
.haskell-ghc-mod.json

# VS
.vscode/

# Emacs
*#
.dir-locals.el
TAGS

# other
.DS_Store
```

PR #34, opened 2020-08-01
Transpose

bolt12:
Useful links:

- https://en.wikipedia.org/wiki/Dual_space#Algebraic_dual_space
- https://en.wikipedia.org/wiki/Dual_basis#Examples
- https://www.youtube.com/watch?v=DA0cnhpSFok
- https://www.youtube.com/watch?v=tzUHJt66ovY
- https://www.youtube.com/watch?v=46AJDYEQ4VA

So I read a lot about linear maps and dual spaces and got something going on. I also left this comments in the Haskell file for easier reference.

---

Suppose `v` and `w` are vector spaces and `T` is a linear transformation between them:

  `T :: L v w s`

  then:
```
          T
       V --> W
        \    |
         \   |
   g . T  \  | g
           \ |
            \|
             S
```

  `g . T` is called `Tº(g)` (read `T` transpose of `g`), so `Tº` interpreted in this way can be seen as having this type:

  `Tº :: V* -> W*`, where `(_)*` is the dual space of a vector space.

  `L v* w* s` is a linear transformation.

  How to translate this into actual code? (See in the actual code)

I gave my best at trying to come up with a solution. I think I'm close but the we still need to figure out if my `trT2` is correct. I'm also not very fond of the `trT2` type signature.

Also in order to implement this I needed to add the `colToL` and `lToCol` dual denotational isomorphism and `colMajorToL` and `lToColMajor` transformations.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -24,7 +24,7 @@ default-extensions:
   # - RankNTypes
   # - StandaloneDeriving
   # - TupleSections
-  # - TypeApplications
+  - TypeApplications
   # - TypeFamilies
   - TypeOperators
   # - UndecidableInstances
--- a/package.yaml
+++ b/package.yaml
@@ -36,10 +36,15 @@ default-extensions:
 
 ghc-options:
   -Wall
+  -Wcompat
+  -Widentities
+  -Wincomplete-uni-patterns
+  -Wredundant-constraints
 
 library:
   dependencies:
     - base
+    - ghc-prim  # GHC.Types (Constraint)
     - distributive
     - adjunctions
   other-modules: []
--- a/src/LinAlg.hs
+++ b/src/LinAlg.hs
@@ -57,6 +57,16 @@ data L :: (* -> *) -> (* -> *) -> (* -> *) where
   JoinL :: V h => h (L f g s) -> L (h :.: f) g s
   ForkL :: V h => h (L f g s) -> L f (h :.: g) s
 
+instance Eq s => Eq (L f g s) where
+  Zero == Zero            = True
+  Zero == _               = False
+  _ == Zero               = False
+  (Scale s) == (Scale s') = s == s'
+  (f :|# g) == (h :| k)   = (f == h) && (g == k)
+  (f :&# g) == (h :& k)   = (f == h) && (g == k)
+  ForkL ms  == Fork ms'   = and $ liftR2 (==) ms ms'
+  JoinL ms  == Join ms'   = and $ liftR2 (==) ms ms'
+
 -- Scalable vectors
 class V a => HasScaleV a where
   scaleV :: Semiring s => s -> L a a s
@@ -71,8 +81,8 @@ instance (HasScaleV a, HasScaleV b) => HasScaleV (a :*: b) where
 instance (HasScaleV a, HasScaleV b, Representable b) => HasScaleV (b :.: a) where
   scaleV s = cross (pureRep (scaleV s))
 
-unjoin2 :: Additive s => L (f :*: g) h s -> L f h s :* L g h s
-unjoin2 Zero = (zero,zero)
+unjoin2 :: L (f :*: g) h s -> L f h s :* L g h s
+unjoin2 Zero = (Zero,Zero)
 unjoin2 (p :|# q) = (p,q)
 unjoin2 ((unjoin2 -> (p,q)) :&# (unjoin2 -> (r,s))) = (p :& r, q :& s)
 unjoin2 (ForkL ms) = (ForkL A.*** ForkL) (unzip (unjoin2 <$> ms))
@@ -91,19 +101,19 @@ unjoin2 (ForkL ms) = (ForkL A.*** ForkL) (unzip (unjoin2 <$> ms))
 (ForkL *** ForkL) (unzip (unjoin <$> ms)) :: L f (k :.: h) s :* L g (k :.: h) s
 #endif
 
-unfork2 :: Additive s => L f (h :*: k) s -> L f h s :* L f k s
-unfork2 Zero = (zero,zero)
+unfork2 :: L f (h :*: k) s -> L f h s :* L f k s
+unfork2 Zero = (Zero,Zero)
 unfork2 (p :&# q) = (p,q)
 unfork2 ((unfork2 -> (p,q)) :|# (unfork2 -> (r,s))) = (p :|# r, q :|# s)
 unfork2 (JoinL ms) = (JoinL A.*** JoinL) (unzip (unfork2 <$> ms))
 
 -- unfork2 ((p :& q) :|# (r :& s)) = (p :|# r, q :|# s)
 
-pattern (:&) :: Additive s => L f h s -> L f k s -> L f (h :*: k) s
+pattern (:&) :: L f h s -> L f k s -> L f (h :*: k) s
 pattern u :& v <- (unfork2 -> (u,v)) where (:&) = (:&#)
 {-# complete (:&) #-}
 
-pattern (:|) :: Additive s => L f h s -> L g h s -> L (f :*: g) h s
+pattern (:|) :: L f h s -> L g h s -> L (f :*: g) h s
 pattern u :| v <- (unjoin2 -> (u,v)) where (:|) = (:|#)
 {-# complete (:|) #-}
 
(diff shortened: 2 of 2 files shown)


bolt12:
Oops realized that I leaved some unused imports that I left by mistake. Also didn't know this comments would show up but they kind of are well placed :stuck_out_tongue:

src/LinAlg.hs
```diff
@@ -155,6 +170,81 @@ diagRep dflt as =
...
+-- Types for linear map transposition
+trT2 f :: L g f s
+
+                      f                     :: L f g s
+                      idL                   :: L (g :.: Par1) (g :.: Par1) s
+                      unforkL idL           :: g (L g Par1 s)
+                      fmap (trT f)          :: g (L g Par1 s) -> g (L g Par1 s) -> g (L f Par1 s)
+       fmap (trT f) $ unforkL idL           :: g (L f Par1 s)
+JoinL $ colToL . lToRow . trT f <$> unforkL :: L g f s
+#endif
```
bolt12, line 234: This is the snippet in question.

From lines 224 to 234 I wanted to show the reasoning behind my head and what the types should look like (the types do not look like that, obviously because of `x :.: Par1`. I ought to remove it since it might confuse things.

bolt12:
Changed the names and added linear map application. See more in #35

src/LinAlg.hs
bolt12: We can also argue a bit about the design space. I chose  `DualSpace v s = L v Par1 s` but I could go for `DualSpace v s = v s` too and use something like `lToCol` or `(@@)` to implement `dual` (previous `trT`)

package.yaml
```diff
@@ -36,6 +36,10 @@ default-extensions:
 
 ghc-options:
   -Wall
+  -Wcompat
+  -Widentities
+  -Wincomplete-uni-patterns
+  -Wredundant-constraints
```
conal, line 42: Thanks for these extra warning options!

src/LinAlg.hs
```diff
@@ -155,6 +173,81 @@ diagRep dflt as =
...
+  @L v* w* s@ is a linear transformation.
+
+  How to translate this into actual code? See below
+
+-}
+
+-- The dual space of @v@ is the set of linear transformations of @v@ to @s@
+-- so a value (term) of this type is a linear transformation over @s@ aka
+-- @L v Par1 s@.
+type DualSpace v s = L v Par1 s
```
conal, line 210: I'd say "`Dual`" instead of "`DualSpace`".

conal: I'd expect instead
``` haskell
dual :: (V2 f g, Semiring s) => L f g s -> L (Dual g) (Dual f) s
```

src/LinAlg.hs
```diff
@@ -155,6 +173,81 @@ diagRep dflt as =
...
+
+-- Knowing this transpose of a linear transformation is just literally the
+-- definition above but structured in a way that makes sense (one could
+-- think that the type of @trT@ should be:
+--
+-- @L f g s -> L (DualSpace g) (DualSpace f) s@
+--
+dual :: (V2 f g, Semiring s) => L f g s -> DualSpace g s -> DualSpace f s
```
conal, line 218: I'd expect instead
``` haskell
dual :: (V2 f g, Semiring s) => L f g s -> L (Dual g) (Dual f) s
```

src/LinAlg.hs
```diff
@@ -155,6 +173,81 @@ diagRep dflt as =
...
+trR :: V2 c r => L c r s -> L r c s
+trR = rowMajToL . distribute . lToRowMaj
+
+-- More elegant, probably efficient version
+tr :: L c r s -> L r c s
+tr (Scale s)  = Scale s
+tr (f :|# g)  = tr f :&# tr g
+tr (f :&# g)  = tr f :|# tr g
+tr (ForkL ms) = JoinL (fmap tr ms)
+tr (JoinL ms) = ForkL (fmap tr ms)
```
conal, line 249: `dual` replaces transposition.

src/LinAlg.hs
```diff
@@ -155,6 +173,81 @@ diagRep dflt as =
 idL :: (V a, Semiring s) => L a a s
 idL = scaleV one
 
+onesV :: (ToScalar a, Representable a, Semiring s) => L a Par1 s
```
conal, line 176: or `Dual a s`.

src/LinAlg.hs
conal: And of course we'll want a version on linear maps/functions that serves to specify this version.

src/LinAlg.hs
```diff
@@ -165,6 +258,14 @@ ForkL ms' .@ m         = ForkL (fmap (.@ m) ms')   -- n-ary product law
 m'        .@ JoinL ms  = JoinL (fmap (m' .@) ms)   -- n-ary coproduct law
 JoinL ms' .@ ForkL ms  = sum (ms' .^ ms)           -- biproduct law
 
+-- Linear map application
+(@@) :: Semiring s => L f g s -> f s -> g s
+Scale s @@ gs             = (* s) <$> gs
+(p :&# q) @@ gs           = p @@ gs :*: q @@ gs
+(p :|# q) @@ (gs :*: gs') = liftR2 (+) (p @@ gs) (q @@ gs')
+ForkL ms @@ gs            = Comp1 (fmap (@@ gs) ms)
+JoinL ms @@ (Comp1 gs)    = foldr (liftR2 (+)) (pureRep zero) (liftR2 (@@) ms gs)
```
conal, line 267: This definition is central to the whole specification, but should probably have the following type instead:

``` haskell
asF :: Semiring s => L f g s -> F f g s
```

where `F` is the type of linear maps/functions, currently defined only in the *specification* branch (and needs a better name).

I’m going to add this function to *specification* and merge. The definition will look quite different but will be equivalent and clearly homomorphic.

BTW, we’ll have to move `s` to the first argument of `F` and `L`, so that `F s` and `L s` can be categories (linear maps over the scalar semiring `s` and a representation of such).

src/LinAlg.hs
```diff
@@ -222,30 +335,53 @@ cross fs == JoinL (ins .^ fs)
...
+khatri a b = (tr fstM .@ a) .* (tr sndM .@ b)
+
 -- See discussion in https://github.com/conal/linalg/pull/26
 
 -- Linear algebra duality isomorphism: a s =~ a s :-* s.
 class ToScalar a where
   rowToL :: a s -> L a Par1 s
+  colToL :: a s -> L Par1 a s
   lToRow :: L a Par1 s -> a s
+  lToCol :: L Par1 a s -> a s
```
conal, line 358: These are the new methods I had in mind, but we’ll either need to move them to a new class “`FromScalar`” or rename this combined class. I see that you did keep `ToColMajor` separate from `ToRowMajor`. By “`ToScalar`” I meant linear maps *to* a scalar, i.e., dual vectors (“single row matrices”). Is there a standard name for linear maps *from* scalars, i.e., `L Par1 g s` (“single column matrices”)? Perhaps these two things are often called “row vectors” and “column vectors” respectively, but doing so seems a confusion between vectors and two types of linear maps isomorphic to vectors.

src/LinAlg.hs
bolt12: Got it!

src/LinAlg.hs
bolt12: Me too but then the implementation wouldn't be as straightforward or faithful to the definition. What do you think I should do?

src/LinAlg.hs
bolt12: Yes, I decided to leave it there because of #35 and I tried to see if the final version would converge to this one in some way. I can remove it or leave it as a comment!

src/LinAlg.hs
bolt12: Also `L` isn't `Representable` so it feels awkward to use it as matrix indices, right?

src/LinAlg.hs
bolt12: I don't know if that alias should be used because it might give the idea that this has a deeper meaning to `onesV` which I don't think is the case

src/LinAlg.hs
bolt12: Why do you say the type signature should be that one? It doesn't make sense to me.. when you apply a linear transformation (`F (f s -> g s)`) you need to pass a `f s`, right? Maybe I'm not understanding correctly, I'll wait to see your implementation!

> BTW, we’ll have to move s to the first argument of F and L, so that F s and L s can be categories (linear maps over the scalar semiring s and a representation of such).

Yes, that makes sense!

src/LinAlg.hs
bolt12: Yes, I can change and create a separate class called `FromScalar`

PR #39, opened 2020-08-05
Matrix Implementation

dwincort:
This is the matrix code I was talking about at the meeting.  I haven't done anything closely with it in the last week or two, which means it doesn't conform to our current coding standards or design philosophy, but hopefully it's readable.

First, I define some category and isomorphism types/classes followed by a class for handling finite cardinality.  This is all almost directly lifted from @conal's previous work.  Then I make an `AbelianCat` class that basically smooshes the idea of semiring and categories into one mess of a class.  (This is sort of a hack because I didn't want to do the work to build up the ideas properly like we're doing in the `specification` branch.).  From there, I introduce the `Matrix` newtype, which is a simple wrapper around `f (g s)` and get right into making strassen work.  The big finale is on line 322 when I write a generic composition function that "pattern matches" on cardinality.

After a brief digression into fork and join notation, I introduce the `MatrixNM` type, which hides `f` and `g` existentially and is indexed instead by the cardinalities of the representations of `f` and `g`.  No real work is done here as everything is performed over isomorphisms to `Matrix`.

I conclude with a few small examples.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 linalg.cabal
+.stack*
+stack.yaml.lock
--- a/package.yaml
+++ b/package.yaml
@@ -46,7 +46,12 @@ library:
     - ghc-prim  # GHC.Types (Constraint)
     - distributive
     - adjunctions
+    - uniform-pair
+    - constraints
+    - vector-sized
+    - finite-typelits
   other-modules: []
   source-dirs: src
   exposed-modules:
     - LinAlg
+    - Matrix
--- a/src/Matrix.hs
+++ b/src/Matrix.hs
@@ -1,0 +1,535 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE NoStarIsType #-}
+{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE PolyKinds #-}
+{-# LANGUAGE Rank2Types #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TupleSections #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE TypeSynonymInstances #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+{-# OPTIONS_GHC -Wno-name-shadowing #-}
+{-# OPTIONS_GHC -Wno-unticked-promoted-constructors #-}
+
+module Matrix where
+
+import Prelude hiding (id, (.))
+
+import qualified Control.Arrow as A
+import qualified Data.Finite   as F
+
+import Control.Category  (Category(..))
+import Data.Bool         (bool)
+import Data.Coerce       (Coercible, coerce)
+import Data.Constraint   ((:-)(..), Dict(..), withDict, (\\))
+import Data.Distributive (Distributive(..))
+import Data.Finite       (Finite)
+import Data.Functor.Rep  (Representable(..), liftR2, pureRep)
+import Data.Kind         (Constraint, Type)
+import Data.Maybe        (fromJust)
+import Data.Proxy        (Proxy(..))
+import Data.UniformPair  (Pair(..))
+import Data.Vector.Sized (Vector)
+import Data.Void         (Void, absurd)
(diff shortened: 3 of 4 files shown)


bolt12:
Awesome!

snowleopard:
Implementing Strassen's algorithm in full generality is a great idea! It's worth noting that it requires a bit more structure from the type of values: they should form a ring (rather than just a semiring), so I expect a `Ring a =>` constraint to pop up at some point.

For showcasing the (star) semiring case, I'd be interested in porting some graph algorithms from my `algebraic-graphs` library. Interestingly, there are implementations that can take advantage of the inductive structure of the matrix, so they might be expressed more naturally via fork/join primitives.

dwincort:
> they should form a ring (rather than just a semiring)

Yes, you're right, and I updated my text above to say "ring" instead of "semiring".  Indeed, the `AbelianCat` type class I define includes a subtraction operation and is definitely more like a ring than a semiring.

conal:
> Implementing Strassen’s algorithm in full generality is a great idea\!

I agree. And I wonder if there’s more generality (and maybe elegance) to be found.

> For showcasing the (star) semiring case, I’d be interested in porting some graph algorithms from my `algebraic-graphs` library. Interestingly, there are implementations that can take advantage of the inductive structure of the matrix, so they might be expressed more naturally via fork/join primitives.

Great. I’m also very interested in applying linear algebra to other interesting semirings, such as optima, languages, and probability distributions, and connections to generalized convolution with the monoid semiring (functions from a monoid to a semiring). The language graphs angle is very appealing, too.

I think the fields of automatic differentiation and deep learning have been held back for decades by defining some wrong choices into their problem statements, chiefly graphs for computations and matrices for linear maps (derivatives). (Also partial derivatives.) Replacing directed graphs by a categorical interface and matrices by a richer variety of linear map representations allowed for vast simplification and generalization. I’d love to collaboratively explore other uses of directed graphs and of matrices looking for similar improvements.

Let’s keep tugging at these threads\!

conal:
> It’s worth noting that it requires a bit more structure from the type of values: they should form a ring (rather than just a semiring), so I expect a `Ring a =>` constraint to pop up at some point.

The ring requirement worries me, since there are a lot of very useful semirings that are not rings. Convolution is also usually defined with subtraction, but there’s a very lovely generalized definition that needs only addition, with subtraction as an optional optimization when available. See [*Generalized Convolution and Efficient Language Recognition*](http://conal.net/papers/convolution/).

dwincort:
> The ring requirement worries me, since there are a lot of very useful semirings that are not rings. Convolution is also usually defined with subtraction, but there’s a very lovely generalized definition that needs only addition, with subtraction as an optional optimization when available. See [_Generalized Convolution and Efficient Language Recognition_](http://conal.net/papers/convolution/).

Perhaps Strassen is the same way?  Maybe it's an optimization that becomes available when subtraction is available.

conal:
> Perhaps Strassen is the same way? Maybe it’s an optimization that becomes available when subtraction is available.

I’m hoping instead that the Strassen algorithm turns out to be a special case of a more general, but still sub-cubic algorithm that doesn’t need subtraction, just as convolution doesn’t. In other words, as with convolution, the generalized form might be much more widely applicable than the specialization that Strassen turns out to be in hindsight while still being quite efficient.

snowleopard:
> I’m hoping instead that the Strassen algorithm turns out to be a special case of a more general, but still sub-cubic algorithm that doesn’t need subtraction, just as convolution doesn’t.

@conal That would be a huge breakthrough in the complexity theory, so I'm sceptical (sorry!). The complexity of matrix multiplication over semirings was studied since 1970s [1]. It is one of the few famous problems in the _subcubic equivalence class_ [2], along with such problems all-pairs shortest paths (APSP), finding a negative triangle, and others. If any of these problems can be solved in subcubic time, then all can. 

I agree with @dwincort here: subtraction appears to be necessary fast matrix multiplication.

P.S.: Plugging my own work in this area, faster algorithms are possible for matrices that are _too dense_, that is, have only O(n) zeroes :-) As we show in [3], a 0/1-matrix with O(n) zeroes can be multiplied by any matrix in O(n^2)!
 
[1] Leslie Kerr. "The effect of algebraic structure on the computation complexity of matrix multiplications", PhD thesis, Cornell University, 1970: https://ecommons.cornell.edu/bitstream/handle/1813/5934/70-75.pdf

[2] Virginia Vassilevska Williams and Ryan Williams. "Subcubic equivalences between path, matrix
and triangle problems", FOCS 2010: https://people.csail.mit.edu/rrw/tria-mmult.pdf

[3] Alexander Kulikov, Ivan Mikhailin, Andrey Mokhov, Vladimir Podolskii. "Complexity of Linear Operators", ISAAC 2019: https://drops.dagstuhl.de/opus/volltexte/2019/11513/pdf/LIPIcs-ISAAC-2019-17.pdf

conal:
> @conal That would be a huge breakthrough in the complexity theory, so I’m sceptical (sorry\!). The complexity of matrix multiplication over semirings was studied since 1970s \[1\]. It is one of the few famous problems in the *subcubic equivalence class* \[2\], along with such problems as all-pairs shortest paths (APSP), finding a negative triangle, and others. If any of these problems can be solved in subcubic time, then all can.

Now I’m much more skeptical as well. Thanks for this info, @snowleopard\!

PR #40, opened 2020-08-08
Better n-ary/representable classes

conal:
Now poly-kinded, more sensibly defined, exhibit the usual product/coproduct dualities, and mirror their binary counterparts. I mostly used [Dan's formulation](https://github.com/conal/linalg/pull/28#discussion_r465142333) but changed the `ComonoidalR` and `CocartesianR` instances of `LinearFunction.L` so that (a) their operations are linear and (b) linear functions form a biproduct category.

### DIFF
--- a/src/Category.hs
+++ b/src/Category.hs
@@ -12,9 +12,11 @@ module Category where
 import qualified Prelude as P
 import Prelude hiding (id,(.))
 import GHC.Types (Constraint)
+import GHC.Generics (Generic,Generic1)
 import qualified Control.Arrow as A
-import GHC.Generics ((:.:)(..))
 import Data.Constraint ((:-)(..),Dict(..),(\\),refl,trans)
+import Data.Distributive
+import Data.Functor.Rep
 
 import Misc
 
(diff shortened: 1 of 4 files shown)


src/Category.hs
```diff
@@ -227,6 +242,36 @@ instance Cocartesian (->) (:+) where
   -- jam (Left  a) = a
   -- jam (Right a) = a
 
+newtype Ap f a = Ap { unAp :: f a }
+  deriving (Functor, Generic, Generic1)
```
dwincort, line 246: You can also use the `Ap` defined in `Data.Monoid`: https://hackage.haskell.org/package/base-4.14.0.0/docs/Data-Monoid.html#t:Ap

conal: Thanks. I’d forgotten about that `Ap`. I’d like to switch, but there don’t seem to be `Distributive` and `Representable` instances, so we’d have to introduce orphan instances. If we encounter any other missing instances, we might add an orphans module and start making pull requests for other libraries.

PR #41, opened 2020-08-09
Category examples

conal:
Moves category examples to new directory, and adds `Isomorphism`. I wanted to declutter the `Category` module while continuing to add simple and useful categories.


conal:
The `Isomorphism` category is more than just another interesting example. Besides being interesting on its own, I hope it will be useful for us in repartitioning linear maps (and their representations), e.g., for using the *Strassen* composition algorithm even when the functor shapes don’t come conveniently packaged. Also for converting between linear map representations, including ones represented by numerically-indexed arrays, as suggested by commit https://github.com/conal/linalg/pull/41/commits/3b6d24c20eb12491f3434d00ebe70deb34c39369 .

src/Category.hs
```diff
@@ -226,6 +288,17 @@ instance Category (->) where
 instance Monoidal (->) (:*) where
   (***) = (A.***)
 
+instance Associative (->) (:*) where
+#ifndef AssociativeDefaultCartesian
+  lassoc (a,(b,c)) = ((a,b),c)
+  rassoc ((a,b),c) = (a,(b,c))
+#endif
```
dwincort, line 295: The use of CPP here is pretty ugly, and I think it can be replaced by using `StandaloneDeriving` along with `DerivingVia`.  That said, I can only get it to work if you switch the order of type arguments in the classes so that `p` comes before `k`.  Consider:
```haskell
class (Category k, ObjBin p k) => Monoidal p k | k -> p where
  ...

class Monoidal p k => Cartesian p k where
  ...

class Associative p k where
  lassoc :: Obj3 k a b c => (a `p` (b `p` c)) `k` ((a `p` b) `p` c)
  rassoc :: Obj3 k a b c => ((a `p` b) `p` c) `k` (a `p` (b `p` c))
  -- No defaults defined.

-- This newtype will act as the "default" instance for `Associative` defaulted with `Cartesian`.
newtype AssociativeCartesian p k a b = AssociativeCartesian (k a b)
deriving instance Category k => Category (AssociativeCartesian p k)
deriving instance (Monoidal p k, ObjBin p (AssociativeCartesian p k)) => Monoidal p (AssociativeCartesian p k)
deriving instance (Cartesian p k, ObjBin p (AssociativeCartesian p k)) => Cartesian p (AssociativeCartesian p k)
instance Cartesian p (AssociativeCartesian p k) => Associative p (AssociativeCartesian p k) where
  lassoc = second exl &&& (exr . exr)
  rassoc = (exl . exl) &&& first  exr
```
With the above machinery in place, we can derive the `Associative` instance for `->`:
```haskell
deriving via (AssociativeCartesian (:*) (->)) instance Associative (:*) (->)
```
Similar newtypes can be defined for `AssociativeCocartesian`, `SymmetricCartesian`, and so on.

src/Category/Index.hs
```diff
@@ -0,0 +1,86 @@
...
+  exr = Indexed exr
+  dup = Indexed dup
+
+instance Comonoidal k (:*:) => Comonoidal (Indexed k) (:+) where
+  Indexed f +++ Indexed g = Indexed (f +++ g)
+
+instance Cocartesian k (:*:) => Cocartesian (Indexed k) (:+) where
+  inl = Indexed inl
+  inr = Indexed inr
+  jam = Indexed jam
```
dwincort, line 70: I mentioned switching the order of `k` and `p` in the previous comment, and it holds up here too.  If they were switched, then I think all of these instances could be replaced with standalone derived instances using `GeneralizedNewtypeDeriving`.

src/Category.hs
conal: I mentioned switching the order of `k` and `p` in the previous comment, and it holds up here too. If they were switched, then I think all of these instances could be replaced with standalone derived instances using `GeneralizedNewtypeDeriving`.

Small comments:

  - Could you replace `ObjBin p (ViaCartesian p k)` and `ObjBin p (ViaCocartesian p k)` by `ObjBin p k` and `ObjBin p k`? Did you consciously choose not to?
  - I’d rename “`p`” to “`co`” in the `Cocartesian` standalone deriving declarations.
  - I’d visually align operators in the “`deriving via`” declarations for easy & reliable comparison.

src/Category/Index.hs
conal: Wonderful\!

How about making a PR? Or, if you prefer, I’ll do the edit myself (referring to URLs for your review comments) and make a few cosmetic changes on the way.

src/Category.hs
conal: >   - Could you replace `ObjBin p (ViaCartesian p k)` and `ObjBin p (ViaCocartesian p k)` by `ObjBin p k` and `ObjBin p k`?

Likewise, replace `Cartesian p (ViaCartesian p k)` and `Cocartesian p (ViaCocartesian p k)` by `Cartesian p k` and `Cocartesian p k` (and rename the second “`p`” to “`co`”)?

src/Category/Index.hs
dwincort: I'd be happy to put the changes into a PR

src/Category.hs
dwincort: > * Could you replace `ObjBin p (ViaCartesian p k)` and `ObjBin p (ViaCocartesian p k)` by `ObjBin p k` and `ObjBin p k`?
> 
> Likewise, replace `Cartesian p (ViaCartesian p k)` and `Cocartesian p (ViaCocartesian p k)` by `Cartesian p k` and `Cocartesian p k` (and rename the second “`p`” to “`co`”)?

I can rename `p` to `co` in the `ViaCocartesian` instances, but the other changes are not so easy.  Once I make the PR, you'll be able to easily poke around at what the type errors are.

dwincort: Oh wait, I figured out what's up.  It turns out that deriving an instance that includes a type synonym doesn't do what I want it to do.

For `ViaCartesian`, I had been generating the `Category` instance like so:
```haskell
deriving instance Category k => Category (ViaCartesian p k)
```
I can't actually figure out what GHC thinks `Obj' (ViaCartesian p k)` _is_, but it's certainly not `Obj k` like we want it to be.  To get around this, I needed to explicitly add `ObjBin p (ViaCartesian p k)` into the context of the other derived instances.  An alternative approach is to manually write the `Category` instance, like so:
```haskell
instance Category k => Category (ViaCartesian p k) where
  type Obj' (ViaCartesian p k) = Obj k
  id = ViaCartesian id
  ViaCartesian g . ViaCartesian f = ViaCartesian (g . f)
```
Now, we can get away with just:
```haskell
deriving instance Monoidal p k => Monoidal p (ViaCartesian p k)
deriving instance Cartesian p k => Cartesian p (ViaCartesian p k)
```

It's frustrating that I can't derive the `Category` instance, but this seems a little nicer overall, so I'll do this in the PR.

conal: Thanks for checking\! Alternatively I think we could move `Obj'` out of `Category`.

src/Category/Index.hs
conal: I just pushed a few more changes. Hopefully won't interact with yours, but you might want to pull the latest just in case.

dwincort: Oops, I was wrong.  I'm in the process of making the PR for swapping `k` and `p` and using `DerivingVia` where applicable, but `Indexed` can't be derived.  Oh well.

src/Category/Isomorphism.hs
```diff
@@ -0,0 +1,133 @@
...
+
+fmapIso :: Functor f => (f <--> g) -> (a -> b) -> (g a -> g b)
+fmapIso fg h = isoFwd fg . fmap h . isoRev fg
+
+-- Don't pattern match fg, since we need two type instantiations.
+-- For instance, the following doesn't type-check:
+--
+--   fmapIso (fg :<-> gf) h = fg . fmap h . gf
+
+-- Is this fmapIso really the most useful and natural variant?
```
dwincort, line 133: In my own code, I came up with a different function of the same name:
```haskell
fmapIso :: Functor f => a <-> b -> f a <-> f b
fmapIso (f :<-> g) = (fmap f :<-> fmap g)
```
Essentially, I'm lifting the idea of an isomorphism between two types through the functor.  I can't quite say which is more useful or natural as they both do different things, but I did make use of my version more than once in my code.

conal: Thanks. Much more sensible looking. I pulled the other weird one from concat, but I don't seem to use it there.

conal:
Everything is back to a sane state, so I'm going to merge.

src/Category/Index.hs
```diff
@@ -0,0 +1,80 @@
...
+
+#if 0
+
+type RepX r i = Rep r :* i
+
+instance MonoidalR r (:.:) k => MonoidalR r RepX (Indexed k) where
+  cross (fmap unIndexed -> fs) = Indexed (cross fs)
+
+-- Hm. Needs unsaturated type synonym. We could make RepX a data type or
+-- newtype, but then it wouldn't be the Rep/index of functor composition.
```
dwincort, line 78: First off, before we make `RepX` a data type or newtype, remember that we already have `RepAnd` defined in `Category.hs` which is exactly that.

> but then it wouldn't be the Rep/index of functor composition

I find this really annoying too, and I can't wait for GHC to get around to approving and merging the unsaturated type family changes.  In the mean time, I wonder if we could brainstorm the concrete repercussions of using `RepAnd` here.  After all, `RepAnd r x` is isomorphic to `(Rep r, x)`.

conal: > First off, before we make `RepX` a data type or newtype, remember that we already have `RepAnd` defined in `Category.hs` which is exactly that.

Agreed. We need it to be a data type or newtype for typing, but we need it to be pair type synonym to match `Rep (r :.: u)`, where `Rep u = i`. I guess we could use a new `(:.:)` isomorphic to the usual one and define `Rep (r :.: u) = RepAnd r (Rep u)`. I don’t think I’d be happy with such a hack.

Another possibility is to try making products and coproducts (etc) be associated type synonyms instead of class parameters. I anticipate a difficult time with type inference, but doing so has the potential of *non-injective* operations. For instance, we could then use natural numbers (not types) as objects, with numeric addition and multiplication, and then you might get the re-association I think you’ve been wanting as *equalities* rather than as isomorphisms. I’m starting to reframe helping the type checker as an interesting and useful degree of freedom rather than just an inconvenience. The technique I like for type-indexed algorithm families gets destroyed, but maybe it will resurface in some other form, hopefully one that a hardware-savvy compiler could exploit. I’m excited about this perspective shift, and I’d like to discuss it on Wednesday.

conal:
BTW, I renamed `Category.Index` to "`Category.Indexed`" in https://github.com/conal/linalg/commit/b668f8d9abde42b21b44f18c8e2cfd7162d58fdf. I meant to make the change in a PR but forgot to create the new branch.

PR #43, opened 2020-08-10
Add custom CatPrelude

dwincort:
As mentioned in [a comment of mine](https://github.com/conal/linalg/pull/28#pullrequestreview-460333676), we can make our own custom prelude so that our imports are a little cleaner.  This PR does this.

For now, I kept it rather simple.  This new `CatPrelude` is basically just the regular `Prelude` with a few modifications:
- `(+)`, `sum`, `(*)`, `unzip`, `id`, and `(.)` are not exported.
- The contents of `Misc` and `Category` are exported.
- The contents of `Data.Distributive` and `Data.Functor.Rep` are exported.
- The pieces of `GHC.Generics` and `GHC.Types` (just `Constraint` in this case) that we use are exported.

Depending on where this repo goes, we could consider more aggressive changes:
- Rely on something like [protolude](https://hackage.haskell.org/package/protolude) instead of GHC's Prelude, which is a modified prelude that, among many nice features, does not include partial functions by default.
- Add more "default" modules, for instance, `Dict` (and related helpers) from `Data.Constraint`.

Note that I specifically didn't force `LinAlg` to use `CatPrelude` because it is not set up for using our `Category` module yet.  I expect that future modules we write (including the category-rewrite of `LinAlg`) will all start with `import CatPrelude`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 linalg.cabal
+.stack*
+stack.yaml.lock
--- a/package.yaml
+++ b/package.yaml
@@ -25,6 +25,7 @@ default-extensions:
   - InstanceSigs
   # - LambdaCase
   - MultiParamTypeClasses
+  - NoImplicitPrelude
   - PatternSynonyms
   - QuantifiedConstraints
   - RankNTypes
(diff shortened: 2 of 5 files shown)


conal:
Lovely! Thanks a bunch.

PR #44, opened 2020-08-10
vector/map isomorphisms

conal:
- `Orphans` module with `Ap` and `Newtype` instances.
- Move `Unconstrained` from `Category` to `Misc`.
- `(:+:)` utilities in `Misc`
- Vector/map isomorphisms in `LinearFunction`, use for inner and outer products

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -57,6 +57,7 @@ library:
   exposed-modules:
     - Misc
     - Category
+    - Orphans
     - LinearFunction
     - InductiveMatrix # coming
     - LinAlg          # going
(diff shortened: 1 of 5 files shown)


conal:
I’m returning to considering various “products” that arise in linear algebra. My current best guesses:

  - Inner: composition of vector-to-scalar linear map with scalar-to-vector linear map.
  - Outer: composition of scalar-to-vector linear map with vector-to-scalar linear map.
  - Matrix/matrix: linear map composition.
  - Kronecker: outer product with linear maps as “vectors”.
  - Khatri-Rao: Kronecker product for linear maps generalized from semirings to abelian categories.
  - Hadamard-Schur: not a composition.

This PR contains a first step toward checking out these guesses. It adds the linear isomorphisms (a) between vectors and vector-to-scalar linear maps, and (b) between vectors and scalar-to-vector linear maps. Composing these two isomorphisms in both orders gives inner and outer products.

PR #45, opened 2020-08-12
Linear map as vector, first try

conal:
A useful fact about linear algebra is that linear maps on vector spaces form a vector space (with all spaces involved sharing a scalar semiring). The dual space notion behind transposition is one example of many. However, with the current interfaces, linear maps have the right kind to be a category (with the scalar type parameter coming first) but the wrong kind to be a vector (where the scalar would come last).

One idea is `newtype`-wrap `L s a b` into `LV a b s`, so `LV a b` (“linear map as vector”) has the kind of a vector, as in this pull request.

Since vectors are representables, we need a `Representable` instance for `LV a b` (and hence `Functor` and `Distributive`), and I think the only reasonable choices are `Rep a :* Rep b` and `Rep b :* Rep a`, corresponding to column-major and column-minor matrices. (We discussed this question in PR \#26.)

As [you will see in this PR](https://github.com/conal/linalg/blob/558a70e0c820eb338efd8784dc9f4fad84752e8a/src/LinearFunction.hs#L126-L142), I ran into a problem that suggests replacing `Representable` with version that constrains the codomain/element type.

A very different tactic is to make linear maps be a category over `*` instead of `* -> *`, but add a constraint that requires an isomorphism between types and representable functors for a given scalar. See section 7.4 and Appendix A of [*Compiling to categories*](http://conal.net/papers/compiling-to-categories) and module [`ConCat.Free.VectorSpace`](https://github.com/conal/concat/blob/master/examples/src/ConCat/Free/VectorSpace.hs) in the implementation. Then `L s a b` could indeed be a vector, and the corresponding representable could be `b :.: a` or `a :.: b`.

Any thoughts on these issues?

### DIFF
--- a/src/LinearFunction.hs
+++ b/src/LinearFunction.hs
@@ -97,6 +97,76 @@ class LinearMap l where
 -- Trivial instance
 instance LinearMap L where mu = id
 
+-- Linear map application (for all representations)
+infixl 9 @@
+(@@) :: (LinearMap l, C2 Representable a b, Obj2 (l s) a b)
+     => l s a b -> a s -> b s
+m @@ a = unL (isoFwd mu m) a
+
+
+-------------------------------------------------------------------------------
+-- | Linear map as vector
+-------------------------------------------------------------------------------
+
+newtype LV l (a :: * -> *) (b :: * -> *) s = LV { unLV :: l s a b }
+
+#if 0
+-- For now instances have l = L, but we will generalize.
+
+instance (Representable a, Eq (Rep a), Foldable a, Representable b)
+      => Functor (LV L a b) where
+  fmap = fmapRep
+
+instance (Representable a, Eq (Rep a), Foldable a, Representable b)
+      => Distributive (LV L a b) where
+  distribute = distributeRep
+
+-- TODO: consider more efficient fmap and distribute definitions
+
+instance (Representable a, Eq (Rep a), Foldable a, Representable b)
+      => Representable (LV L a b) where
+  type Rep (LV L a b) = Rep b :* Rep a
+  tabulate = LV . unMat . tabulate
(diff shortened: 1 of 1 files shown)


bolt12:
Interesting! With this we can have the sensible type signature of the `dual` function being `dual :: L s a b -> L s (LV b Par1) (LV a Par1)`, right? 

But I'm worried about having to define a custom `Representable` instance just because it would mean redoing a lot of work.

P.S: I've been trying to keep up with the latest developments but the paper presentation is consuming more time, so I apologise for my absence in these upcoming days. The current developments are outstanding, I think this is going in an interesting direction! :smiley:

bolt12:
In order to give a `Representable` instance `LV` needs to be an instance of `Generic1` and others:

```haskell
instance (Representable a, Eq (Rep a), Foldable a, Representable b)
      => Representable (LV L a b) where
  type Rep (LV L a b) = Rep b :* Rep a
  -- tabulate = LV . unMat . tabulate
  -- index = index . toMat . unLV
```

image

conal:
> Interesting\! With this we can have the sensible type signature of the `dual` function being `dual :: L s a b -> L s (LV L b Par1) (LV L a Par1)`, right?

Yes, and hopefully more generally, for *every* linear map representation (one definition for all, including representations still to be defined).

> P.S: I’ve been trying to keep up with the latest developments but the paper presentation is consuming more time, so I apologise for my absence in these upcoming days. The current developments are outstanding, I think this is going in an interesting direction\! :smiley:

No problem. Good luck with your presentation, and congratulations\!

bolt12:
Thank you @conal !

I apologise for the unstructured comments above but I was editing and refactoring them as I tried different things. I concluded that a custom `Representable` type class.

PR #51, opened 2020-08-16
Reordered (.@) clauses to reduce duplication/work.

conal:
I'm giving @bolt12's recent *[Type Your Matrices](https://github.com/bolt12/tymfgg-pearl)* paper a closer read and noticed the clause order for `comp` in section 4. Now I realize that the order I was using caused avoidable duplication.


conal:
Oh, hm. On second thought, I guess these clauses are non-overlapping, so reordering doesn't change anything algorithmic.

bolt12:
Yes I also think that there's no problem with the clause order!

conal:
Thanks. I'm closing this one without merging.

PR #52, opened 2020-08-16
Make the `Obj'` type family depend on object as well as category.

conal:
The `Category` class in section 6 of [*Type Your Matrices*](https://github.com/bolt12/tymfgg-pearl) uses an associated constraint that depends on the object as well as the category. Mine depended only on the category and so sometimes necessitates defining new classes. I carried over this choice from concat, but I guess the motivation has gone away with the introduction of quantified constraints or some other change. (Now I wonder if concat even needs it. In either case, I should probably switch concat over to quantified constraints as well and eliminate all of the constraint entailment programming.) I suspect it helps considerably that we define an [extra class that wraps the associated `Obj'` constraint](https://github.com/conal/linalg/pull/28#issuecomment-670313952).

This change eliminated the need for `Unconstrained`, so I dropped it.

### DIFF
--- a/src/Category.hs
+++ b/src/Category.hs
@@ -32,8 +32,8 @@ instance Obj' k a => Obj k a
 -- Use UndecidableSuperClasses to accept this
 
 class Category (k :: u -> u -> *) where
-  type Obj' k :: u -> Constraint
-  type instance Obj' k = Unconstrained
+  type Obj' k (a :: u) :: Constraint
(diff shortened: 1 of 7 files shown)


bolt12:
For the record I believe I took inspiration on this library: https://hackage.haskell.org/package/constrained-categories-0.4.0.0/docs/Control-Category-Constrained.html

But this does simplifies things a bit!

PR #53, opened 2020-08-16
Unify Comonoidal with Monoidal

conal:
Unify `Comonoidal` with `Monoidal`

*   Drop functional dependencies in associated product, coproduct, and exponential.
*   Unify `Comonoidal` with `Monoidal` and `ComonoidalR` with `MonoidalR`.
*   Merge `(***)` and `(+++)` to become `(###)`.
    Merge `cross` and `plus` to become `bifunctor`.
*   Drop some `UndecidableInstances` pragmas.

### DIFF
--- a/src/Category.hs
+++ b/src/Category.hs
@@ -55,7 +55,7 @@ type Obj6 k a b c d e f = C6 (Obj k) a b c d e f
 -- Seee https://github.com/conal/linalg/pull/28#issuecomment-670313952
 type ObjBin p k = ((forall a b. Obj2 k a b => Obj k (a `p` b)) :: Constraint)
 
-class (Category k, ObjBin p k) => Monoidal p k | k -> p where
+class (Category k, ObjBin p k) => Monoidal p k {- | k -> p -} where
   infixr 3 ***
   (***) :: Obj4 k a b c d => (a `k` c) -> (b `k` d) -> ((a `p` b) `k` (c `p` d))
 
@@ -141,7 +141,7 @@ class Symmetric p k where
 -- <https://hackage.haskell.org/package/categories/docs/Control-Category-Monoidal.html>.
 
 
-class (Category k, ObjBin co k) => Comonoidal co k | k -> co where
+class (Category k, ObjBin co k) => Comonoidal co k {- | k -> co -} where
   infixr 2 +++
   (+++) :: Obj4 k a b c d => (a `k` c) -> (b `k` d) -> ((a `co` b) `k` (c `co` d))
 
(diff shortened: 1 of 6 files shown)


conal:
I had thought the functional dependencies necessary for type inference, but they weren’t needed for `Associative` and `Symmetric`, and it turns not for `Monoidal`, `Comonoidal`, `MonoidalR`, and `ComonoidalR` either. With the functional dependencies gone, `Monoidal` and `MonoidalR` accommodates both products and coproducts (just as `Associative` and `Symmetric` already did).

I’m not thrilled with the names “`(###)`” and “`bifunctor`”.

bolt12:
I quite like `(###)` but agree that `bifunctor` isn't as nice.

dwincort:
> I had thought the functional dependencies necessary for type inference, but they weren’t needed

This is great!  It's really cool that type inference still works effectively even with so much overlap.

As for the names, I wonder if something to do with `lift` or `sequence` makes sense instead of `bifunctor`.

conal:
I just noticed that [*Type Your Matrices*](https://github.com/bolt12/tymfgg-pearl) uses `bimap`, which I like.

conal:
> I just noticed that [*Type Your Matrices*](https://github.com/bolt12/tymfgg-pearl) uses `bimap`, which I like.

Oh, never mind. The `bimap` in TYM is `(###)`.

But maybe “`nmap`” or “`rmap`” or “`mapR`”.

conal:
> > I just noticed that [*Type Your Matrices*](https://github.com/bolt12/tymfgg-pearl) uses `bimap`, which I like.
> 
> Oh, never mind. The `bimap` in TYM is `(###)`.

Wrong again. I saw `bimap` in [*Selective Applicative Functors*](https://www.staff.ncl.ac.uk/andrey.mokhov/selective-functors.pdf). Different thing with a teasingly similar type:
``` haskell
bimap :: (a -> c) -> (b -> d) -> Either a b -> Either c d
```

Compare with
``` haskell
(-|-) :: (...) => Matrix e n k -> Matrix e m j -> Matrix e (Either n m) (Either k j)
```

conal:
I think the categorical generalization of `bimap` for a monoidal category `(~>)` with product or coproduct `(⊗)` is

``` haskell
bimap :: (a ~> c) -> (b ~> d) -> (a ⊗ b ~> c ⊗ d)
```

So I guess “`bifunctor`” (previously `cross` and `plus`) is the n-ary version, and a name like “`nmap`”, “`rmap`”, “`mapR`” may fit after all. We could define `bimap2 = uncurry (###)`, akin to `join2 = uncurry (|||)` and `fork2 = uncurry (&&&)`. I guess I didn’t give names to `uncurry (***)` and `uncurry (+++)` before because their not isomorphisms (in contrast to `join2` and `fork2`.

bolt12:
I like `nmap` and `rmap` as well ! `mapR` is a lot similar to `Functor`'s `map`.

conal:
Now I realize that the “bi” is “`bifunctor`” doesn’t fit.

PR #54, opened 2020-08-19
InductiveMatrix

bolt12:
This PR is a first attempt to fill InductiveMatrix.

I'm not sure if this is how I am supposed to do it but I gave it a shot!

### DIFF
--- a/src/Category.hs
+++ b/src/Category.hs
@@ -13,6 +13,7 @@ import GHC.Types (Constraint)
 import qualified Control.Arrow as A
 import Data.Monoid (Ap(..))
 import Data.Functor.Rep
+import GHC.Generics ((:.:)(..))
 
 import Misc
 import Orphans ()
@@ -249,6 +250,10 @@ fork fs = rmap fs . dups
 unfork :: (CartesianR r p k, Obj2 k a b) => a `k` (p r b) -> r (a `k` b)
 unfork f = (. f) <$> exs
 
+pattern Fork :: (Obj' k f, Obj' k g, CartesianR h (:.:) k) => h (k f g) -> k f (h :.: g)
+pattern Fork ms <- (unfork -> ms) where Fork = fork
+-- {-# complete Fork #-} -- See (:&) above
+
 -- Exercise: Prove that fork and unfork form an isomorphism.
 
 -- N-ary biproducts
@@ -261,6 +266,10 @@ join fs = jams . rmap fs
 
 unjoin :: (CocartesianR r co k, Obj2 k a b) => co r a `k` b -> r (a `k` b)
 unjoin f = (f .) <$> ins
+
+pattern Join :: (Obj' k f, Obj' k g, CocartesianR h (:.:) k) => h (k f g) -> k (h :.: f) g
+pattern Join ms <- (unjoin -> ms) where Join = join
+-- {-# complete Join #-} -- See (:&) above
 
 -- TODO: Add fork & unfork to CartesianR with the current definitions as
 -- defaults, and give defaults for exs and dups in terms of fork and unfork.
--- a/src/Category.hs
+++ b/src/Category.hs
@@ -170,7 +170,6 @@ type Bicartesian p co k = (Cartesian p k, Cocartesian co k)
 -- because there are more laws.
 class Bicartesian p p k => Biproduct p k
 
-
 class (Category k, ObjBin e k) => Closed e k where
   (^^^) :: Obj4 k a b c d => (a `k` b) -> (d `k` c) -> ((c `e` a) `k` (d `e` b))
 
@@ -197,8 +196,10 @@ instance Category k => Category (ViaCartesian p k) where
   type Obj' (ViaCartesian p k) a = Obj k a
   id = ViaCartesian id
   ViaCartesian g . ViaCartesian f = ViaCartesian (g . f)
-instance (Category k, Cartesian p (ViaCartesian p k)) => Monoidal p (ViaCartesian p k) where
+instance Cartesian p k => Monoidal p (ViaCartesian p k) where
   f ### g = (f . exl) &&& (g . exr)
+instance (Category k, Representable r, CartesianR r p (ViaCartesian p k)) => MonoidalR r p (ViaCartesian p k) where
+  rmap fs = fork (liftR2 (.) fs exs)
 deriving instance Cartesian p k => Cartesian p (ViaCartesian p k)
 
 instance Cartesian p k => Associative p (ViaCartesian p k) where
@@ -214,8 +215,10 @@ instance Category k => Category (ViaCocartesian p k) where
   type Obj' (ViaCocartesian p k) a = Obj k a
   id = ViaCocartesian id
   ViaCocartesian g . ViaCocartesian f = ViaCocartesian (g . f)
-instance (Category k, Cocartesian co (ViaCocartesian co k)) => Monoidal co (ViaCocartesian co k) where
+instance Cocartesian co k => Monoidal co (ViaCocartesian co k) where
   f ### g = (inl . f) ||| (inr . g)
+instance (Category k, Representable r, CocartesianR r p (ViaCocartesian p k)) => MonoidalR r p (ViaCocartesian p k) where
+  rmap fs = join (liftR2 (.) ins fs)
 deriving instance Cocartesian co k => Cocartesian co (ViaCocartesian co k)
 
 instance Cocartesian co k => Associative co (ViaCocartesian co k) where
--- a/src/Category.hs
+++ b/src/Category.hs
@@ -83,30 +83,24 @@ second :: (Monoidal p k, Obj3 k a b d) => (b `k` d) -> ((a `p` b) `k` (a `p` d))
 second g = id ### g
 
 class Monoidal p k => Cartesian p k where
+  {-# MINIMAL exl, exr #-}
   exl :: Obj2 k a b => (a `p` b) `k` a
   exr :: Obj2 k a b => (a `p` b) `k` b
   dup :: Obj  k a   => a `k` (a `p` a)
-
--- Binary fork
-infixr 3 &&&
-(&&&) :: (Cartesian p k, Obj3 k a c d)
-      => (a `k` c) -> (a `k` d) -> (a `k` (c `p` d))
-f &&& g = (f ### g) . dup
+  dup = id &&& id
+  infixr 3 &&&
+  -- Binary fork
+  (&&&) :: Obj3 k a c d => (a `k` c) -> (a `k` d) -> (a `k` (c `p` d))
+  f &&& g = (f ### g) . dup
+  -- Inverse of fork2
+  unfork2 :: Obj3 k a c d => (a `k` (c `p` d)) -> ((a `k` c) :* (a `k` d))
+  unfork2 f = (exl . f , exr . f)
 
 fork2 :: (Cartesian p k, Obj3 k a c d)
       => (a `k` c) :* (a `k` d) -> (a `k` (c `p` d))
 fork2 = uncurry (&&&)
 
--- Inverse of fork2
-unfork2 :: (Cartesian p k, Obj3 k a c d)
-        => (a `k` (c `p` d)) -> ((a `k` c) :* (a `k` d))
-unfork2 f = (exl . f , exr . f)
-
 -- Exercise: Prove that uncurry (&&&) and unfork2 form an isomorphism.
-
--- TODO: Add (&&&) and unfork2 to Cartesian with the current definitions as
--- defaults, and give defaults for exl, exr, and dup in terms of (&&&) and
--- unfork2. Use MINIMAL pragmas.
 
 pattern (:&) :: (Cartesian p k, Obj3 k a c d)
              => (a `k` c) -> (a `k` d) -> (a `k` (c `p` d))
@@ -119,6 +113,10 @@ pattern f :& g <- (unfork2 -> (f,g)) where (:&) = (&&&)
 --   In {-# complete :& #-}
 --
 -- Instead, give a typed COMPLETE pragma with each cartesian category instance.
+--
+-- Oops! GHC objects to these more specific 'COMPLETE' pragmas (in modules
+-- importing this one) as "orphans" and rejects them. Until this GHC situation
+-- improves, we'll have to <find another solution>.
 
 class Associative p k where
   lassoc :: Obj3 k a b c => (a `p` (b `p` c)) `k` ((a `p` b) `p` c)
@@ -134,30 +132,24 @@ class Symmetric p k where
 -- <https://hackage.haskell.org/package/categories/docs/Control-Category-Monoidal.html>.
 
 class Monoidal co k => Cocartesian co k where
+  {-# MINIMAL inl, inr #-}
   inl :: Obj2 k a b => a `k` (a `co` b)
   inr :: Obj2 k a b => b `k` (a `co` b)
   jam :: Obj  k a   => (a `co` a) `k` a
-
--- Binary join
-infixr 2 |||
-(|||) :: (Cocartesian co k, Obj3 k a b c)
-      => (a `k` c) -> (b `k` c) -> ((a `co` b) `k` c)
-f ||| g = jam . (f ### g)
+  jam = id ||| id
+  -- Binary join
+  infixr 2 |||
+  (|||) :: Obj3 k a b c => (a `k` c) -> (b `k` c) -> ((a `co` b) `k` c)
+  f ||| g = jam . (f ### g)
+  -- Inverse of join2
+  unjoin2 :: Obj3 k a b c => ((a `co` b) `k` c) -> ((a `k` c) :* (b `k` c))
+  unjoin2 f = (f . inl , f . inr)
 
 join2 :: (Cocartesian co k, Obj3 k a b c)
       => (a `k` c) :* (b `k` c) -> ((a `co` b) `k` c)
 join2 = uncurry (|||)
 
--- Inverse of join2
-unjoin2 :: (Cocartesian co k, Obj3 k a b c)
-        => ((a `co` b) `k` c) -> ((a `k` c) :* (b `k` c))
-unjoin2 f = (f . inl , f . inr)
-
 -- Exercise: Prove that uncurry (|||) and unjoin2 form an isomorphism.
-
--- TODO: Add (|||) and unjoin2 to Cartesian with the current definitions as
--- defaults, and give defaults for exl, exr, and dup in terms of (|||) and
--- unjoin2. Use MINIMAL pragmas.
 
 pattern (:|) :: (Cocartesian co k, Obj3 k a b c)
              => (a `k` c) -> (b `k` c) -> ((a `co` b) `k` c)
@@ -237,7 +229,7 @@ instance Cocartesian co k => Symmetric co (ViaCocartesian co k) where
 -- | n-ary counterparts (where n is a type, not a number).
 -------------------------------------------------------------------------------
 
--- Assumes functor categories. To do: look for a clean, poly-kinded alternative.
+-- Assumes functor categories. TODO: look for a clean, poly-kinded alternative.
 -- I guess we could generalize from functor composition and functor application.
 
 type ObjR' r p k = ((forall z. Obj k z => Obj k (p r z)) :: Constraint)
@@ -250,14 +242,13 @@ class (Category k, ObjR r p k) => MonoidalR r p k where
   rmap :: Obj2 k a b => r (a `k` b) -> (p r a `k` p r b)
 
 class MonoidalR r p k => CartesianR r p k where
+  {-# MINIMAL exs, dups #-}
   exs  :: Obj k a => r (p r a `k` a)
   dups :: Obj k a => a `k` p r a
-
-fork :: (CartesianR r p k, Obj2 k a c) => r (a `k` c) -> (a `k` p r c)
-fork fs = rmap fs . dups
-
-unfork :: (CartesianR r p k, Obj2 k a b) => a `k` (p r b) -> r (a `k` b)
-unfork f = (. f) <$> exs
+  fork :: Obj2 k a c => r (a `k` c) -> (a `k` p r c)
+  fork fs = rmap fs . dups
+  unfork :: Obj2 k a b => a `k` (p r b) -> r (a `k` b)
+  unfork f = (. f) <$> exs
 
 pattern Fork :: (CartesianR h p k, Obj2 k f g) => h (k f g) -> k f (p h g)
 pattern Fork ms <- (unfork -> ms) where Fork = fork
@@ -267,22 +258,17 @@ pattern Fork ms <- (unfork -> ms) where Fork = fork
 
 -- N-ary biproducts
 class MonoidalR r co k => CocartesianR r co k where
+  {-# MINIMAL ins, jams #-}
   ins  :: Obj k a => r (a `k` co r a)
   jams :: Obj k a => co r a `k` a
-
-join :: (CocartesianR r co k, Obj2 k a b) => r (a `k` b) -> co r a `k` b
-join fs = jams . rmap fs
-
-unjoin :: (CocartesianR r co k, Obj2 k a b) => co r a `k` b -> r (a `k` b)
-unjoin f = (f .) <$> ins
+  join :: Obj2 k a b => r (a `k` b) -> co r a `k` b
+  join fs = jams . rmap fs
+  unjoin :: Obj2 k a b => co r a `k` b -> r (a `k` b)
+  unjoin f = (f .) <$> ins
 
 pattern Join :: (CocartesianR h p k, Obj2 k f g) => h (k f g) -> k (p h f) g
 pattern Join ms <- (unjoin -> ms) where Join = join
 -- {-# complete Join #-} -- See (:&) above
-
--- TODO: Add fork & unfork to CartesianR with the current definitions as
--- defaults, and give defaults for exs and dups in terms of fork and unfork.
--- Ditto for ins/jams and join/unjoin. Use MINIMAL pragmas.
 
 type BicartesianR r p co k = (CartesianR r p k, CocartesianR r co k)
 
--- a/src/Category.hs
+++ b/src/Category.hs
@@ -198,13 +198,14 @@ instance Category k => Category (ViaCartesian p k) where
   type Obj' (ViaCartesian p k) a = Obj k a
   id = ViaCartesian id
   ViaCartesian g . ViaCartesian f = ViaCartesian (g . f)
-deriving instance Monoidal  p k => Monoidal  p (ViaCartesian p k)
-deriving instance Cartesian p k => Cartesian p (ViaCartesian p k)
+instance (Category k, Cocartesian p (ViaCartesian p k)) => Monoidal p (ViaCartesian p k) where
+  f ### g = (inl . f) ||| (inr . g)
+deriving instance (Cartesian p k, Cocartesian p (ViaCartesian p k)) => Cartesian p (ViaCartesian p k)
 
-instance Cartesian p k => Associative p (ViaCartesian p k) where
+instance (Cartesian p k, Cocartesian p (ViaCartesian p k)) => Associative p (ViaCartesian p k) where
   lassoc = second exl &&& (exr . exr)
   rassoc = (exl . exl) &&& first  exr
-instance Cartesian p k => Symmetric p (ViaCartesian p k) where
+instance (Cartesian p k, Cocartesian p (ViaCartesian p k)) => Symmetric p (ViaCartesian p k) where
   swap = exr &&& exl
 
 -- | The 'ViaCocartesian' type is designed to be used with `DerivingVia` to derive
@@ -214,15 +215,15 @@ instance Category k => Category (ViaCocartesian p k) where
   type Obj' (ViaCocartesian p k) a = Obj k a
   id = ViaCocartesian id
   ViaCocartesian g . ViaCocartesian f = ViaCocartesian (g . f)
-deriving instance Monoidal  co k => Monoidal  co (ViaCocartesian co k)
-deriving instance Cocartesian co k => Cocartesian co (ViaCocartesian co k)
+instance (Category k, Cartesian co (ViaCocartesian co k)) => Monoidal co (ViaCocartesian co k) where
+  f ### g = (f . exl) &&& (g . exr)
+deriving instance (Cartesian co (ViaCocartesian co k), Cocartesian co k) => Cocartesian co (ViaCocartesian co k)
 
-instance Cocartesian co k => Associative co (ViaCocartesian co k) where
+instance (Cartesian co (ViaCocartesian co k), Cocartesian co k) => Associative co (ViaCocartesian co k) where
   lassoc = inl.inl ||| (inl.inr ||| inr)
   rassoc = (inl ||| inr.inl) ||| inr.inr
-instance Cocartesian co k => Symmetric co (ViaCocartesian co k) where
+instance (Cartesian co (ViaCocartesian co k), Cocartesian co k) => Symmetric co (ViaCocartesian co k) where
   swap = inr ||| inl
-
 
 -------------------------------------------------------------------------------
 -- | n-ary counterparts (where n is a type, not a number).
@@ -318,10 +319,10 @@ instance MonoidalClosed (:*) (->) (->) where
   curry   = P.curry
   uncurry = P.uncurry
 
-deriving via (ViaCartesian   (:*) (->)) instance Associative (:*) (->)
-deriving via (ViaCartesian   (:*) (->)) instance Symmetric   (:*) (->)
-deriving via (ViaCocartesian (:+) (->)) instance Associative (:+) (->)
-deriving via (ViaCocartesian (:+) (->)) instance Symmetric   (:+) (->)
+deriving via (ViaCartesian   (:*) (->)) instance (Cocartesian (,) (ViaCartesian (,) (->))) => Associative (:*) (->)
+deriving via (ViaCartesian   (:*) (->)) instance (Cocartesian (,) (ViaCartesian (,) (->))) => Symmetric   (:*) (->)
+deriving via (ViaCocartesian (:+) (->)) instance (Cartesian Either (ViaCocartesian Either (->))) => Associative (:+) (->)
+deriving via (ViaCocartesian (:+) (->)) instance (Cartesian Either (ViaCocartesian Either (->))) => Symmetric   (:+) (->)
 
 instance Representable r => MonoidalR r Ap (->) where
   rmap rab (Ap ra) = Ap (liftR2 ($) rab ra)
--- a/src/Category.hs
+++ b/src/Category.hs
@@ -83,9 +83,11 @@ second :: (Monoidal p k, Obj3 k a b d) => (b `k` d) -> ((a `p` b) `k` (a `p` d))
 second g = id ### g
 
 class Monoidal p k => Cartesian p k where
-  {-# MINIMAL exl, exr #-}
+  {-# MINIMAL exl, exr, ((&&&) | dup) #-}
   exl :: Obj2 k a b => (a `p` b) `k` a
+  exl = fst $ unfork2 id
   exr :: Obj2 k a b => (a `p` b) `k` b
+  exr = snd $ unfork2 id
   dup :: Obj  k a   => a `k` (a `p` a)
   dup = id &&& id
   infixr 3 &&&
@@ -132,9 +134,11 @@ class Symmetric p k where
 -- <https://hackage.haskell.org/package/categories/docs/Control-Category-Monoidal.html>.
 
 class Monoidal co k => Cocartesian co k where
-  {-# MINIMAL inl, inr #-}
+  {-# MINIMAL inl, inr, ((|||) | jam) #-}
   inl :: Obj2 k a b => a `k` (a `co` b)
+  inl = fst $ unjoin2 id
   inr :: Obj2 k a b => b `k` (a `co` b)
+  inr = snd $ unjoin2 id
   jam :: Obj  k a   => (a `co` a) `k` a
   jam = id ||| id
   -- Binary join
--- a/src/Category.hs
+++ b/src/Category.hs
@@ -83,7 +83,7 @@ second :: (Monoidal p k, Obj3 k a b d) => (b `k` d) -> ((a `p` b) `k` (a `p` d))
 second g = id ### g
 
 class Monoidal p k => Cartesian p k where
-  {-# MINIMAL exl, exr, ((&&&) | dup) #-}
+  {-# MINIMAL ((exl, exr) | unfork2), ((&&&) | dup) #-}
   exl :: Obj2 k a b => (a `p` b) `k` a
   exl = fst $ unfork2 id
   exr :: Obj2 k a b => (a `p` b) `k` b
@@ -134,7 +134,7 @@ class Symmetric p k where
 -- <https://hackage.haskell.org/package/categories/docs/Control-Category-Monoidal.html>.
 
 class Monoidal co k => Cocartesian co k where
-  {-# MINIMAL inl, inr, ((|||) | jam) #-}
+  {-# MINIMAL ((inl, inr) | unjoin2), ((|||) | jam) #-}
   inl :: Obj2 k a b => a `k` (a `co` b)
   inl = fst $ unjoin2 id
   inr :: Obj2 k a b => b `k` (a `co` b)
@@ -238,20 +238,22 @@ instance Cocartesian co k => Symmetric co (ViaCocartesian co k) where
 
 type ObjR' r p k = ((forall z. Obj k z => Obj k (p r z)) :: Constraint)
 
-class    (Functor r, ObjR' r p k) => ObjR r p k
-instance (Functor r, ObjR' r p k) => ObjR r p k
+class    (Representable r, ObjR' r p k) => ObjR r p k
+instance (Representable r, ObjR' r p k) => ObjR r p k
 
 class (Category k, ObjR r p k) => MonoidalR r p k where
   -- | n-ary version of '(###)'
   rmap :: Obj2 k a b => r (a `k` b) -> (p r a `k` p r b)
 
 class MonoidalR r p k => CartesianR r p k where
-  {-# MINIMAL exs, dups #-}
+  {-# MINIMAL ((exs | unfork), (fork | dups)) #-}
   exs  :: Obj k a => r (p r a `k` a)
+  exs = unfork id
   dups :: Obj k a => a `k` p r a
+  dups = fork (pureRep id)
   fork :: Obj2 k a c => r (a `k` c) -> (a `k` p r c)
   fork fs = rmap fs . dups
-  unfork :: Obj2 k a b => a `k` (p r b) -> r (a `k` b)
+  unfork :: Obj2 k a b => a `k` p r b -> r (a `k` b)
   unfork f = (. f) <$> exs
 
 pattern Fork :: (CartesianR h p k, Obj2 k f g) => h (k f g) -> k f (p h g)
@@ -262,9 +264,11 @@ pattern Fork ms <- (unfork -> ms) where Fork = fork
 
 -- N-ary biproducts
 class MonoidalR r co k => CocartesianR r co k where
-  {-# MINIMAL ins, jams #-}
+  {-# MINIMAL ((ins | unjoin), (join | jams)) #-}
   ins  :: Obj k a => r (a `k` co r a)
+  ins = unjoin id
   jams :: Obj k a => co r a `k` a
+  jams = join (pureRep id)
   join :: Obj2 k a b => r (a `k` b) -> co r a `k` b
   join fs = jams . rmap fs
   unjoin :: Obj2 k a b => co r a `k` b -> r (a `k` b)
--- a/src/Category.hs
+++ b/src/Category.hs
@@ -198,9 +198,12 @@ instance Category k => Category (ViaCartesian p k) where
   ViaCartesian g . ViaCartesian f = ViaCartesian (g . f)
 instance Cartesian p k => Monoidal p (ViaCartesian p k) where
   f ### g = (f . exl) &&& (g . exr)
-instance (Category k, Representable r, CartesianR r p (ViaCartesian p k)) => MonoidalR r p (ViaCartesian p k) where
+instance (Representable r, CartesianR r p k) => MonoidalR r p (ViaCartesian p k) where
   rmap fs = fork (liftR2 (.) fs exs)
 deriving instance Cartesian p k => Cartesian p (ViaCartesian p k)
+instance (CartesianR r p k, Representable r) => CartesianR r p (ViaCartesian p k) where
+  exs = ViaCartesian <$> exs
+  dups = ViaCartesian dups
 
 instance Cartesian p k => Associative p (ViaCartesian p k) where
   lassoc = second exl &&& (exr . exr)
@@ -217,9 +220,12 @@ instance Category k => Category (ViaCocartesian p k) where
   ViaCocartesian g . ViaCocartesian f = ViaCocartesian (g . f)
 instance Cocartesian co k => Monoidal co (ViaCocartesian co k) where
   f ### g = (inl . f) ||| (inr . g)
-instance (Category k, Representable r, CocartesianR r p (ViaCocartesian p k)) => MonoidalR r p (ViaCocartesian p k) where
+instance (CocartesianR r p k, Representable r) => MonoidalR r p (ViaCocartesian p k) where
   rmap fs = join (liftR2 (.) ins fs)
 deriving instance Cocartesian co k => Cocartesian co (ViaCocartesian co k)
+instance (CocartesianR r p k, Representable r) => CocartesianR r p (ViaCocartesian p k) where
+  ins = ViaCocartesian <$> ins
+  jams = ViaCocartesian jams
 
 instance Cocartesian co k => Associative co (ViaCocartesian co k) where
   lassoc = inl.inl ||| (inl.inr ||| inr)
--- a/src/Category.hs
+++ b/src/Category.hs
@@ -13,7 +13,6 @@ import GHC.Types (Constraint)
 import qualified Control.Arrow as A
 import Data.Monoid (Ap(..))
 import Data.Functor.Rep
-import GHC.Generics ((:.:)(..))
 
 import Misc
 import Orphans ()
@@ -198,14 +197,14 @@ instance Category k => Category (ViaCartesian p k) where
   type Obj' (ViaCartesian p k) a = Obj k a
   id = ViaCartesian id
   ViaCartesian g . ViaCartesian f = ViaCartesian (g . f)
-instance (Category k, Cocartesian p (ViaCartesian p k)) => Monoidal p (ViaCartesian p k) where
-  f ### g = (inl . f) ||| (inr . g)
-deriving instance (Cartesian p k, Cocartesian p (ViaCartesian p k)) => Cartesian p (ViaCartesian p k)
+instance (Category k, Cartesian p (ViaCartesian p k)) => Monoidal p (ViaCartesian p k) where
+  f ### g = (f . exl) &&& (g . exr)
+deriving instance Cartesian p k => Cartesian p (ViaCartesian p k)
 
-instance (Cartesian p k, Cocartesian p (ViaCartesian p k)) => Associative p (ViaCartesian p k) where
+instance Cartesian p k => Associative p (ViaCartesian p k) where
   lassoc = second exl &&& (exr . exr)
   rassoc = (exl . exl) &&& first  exr
-instance (Cartesian p k, Cocartesian p (ViaCartesian p k)) => Symmetric p (ViaCartesian p k) where
+instance Cartesian p k => Symmetric p (ViaCartesian p k) where
   swap = exr &&& exl
 
 -- | The 'ViaCocartesian' type is designed to be used with `DerivingVia` to derive
@@ -215,14 +214,14 @@ instance Category k => Category (ViaCocartesian p k) where
   type Obj' (ViaCocartesian p k) a = Obj k a
   id = ViaCocartesian id
   ViaCocartesian g . ViaCocartesian f = ViaCocartesian (g . f)
-instance (Category k, Cartesian co (ViaCocartesian co k)) => Monoidal co (ViaCocartesian co k) where
-  f ### g = (f . exl) &&& (g . exr)
-deriving instance (Cartesian co (ViaCocartesian co k), Cocartesian co k) => Cocartesian co (ViaCocartesian co k)
+instance (Category k, Cocartesian co (ViaCocartesian co k)) => Monoidal co (ViaCocartesian co k) where
+  f ### g = (inl . f) ||| (inr . g)
+deriving instance Cocartesian co k => Cocartesian co (ViaCocartesian co k)
 
-instance (Cartesian co (ViaCocartesian co k), Cocartesian co k) => Associative co (ViaCocartesian co k) where
+instance Cocartesian co k => Associative co (ViaCocartesian co k) where
   lassoc = inl.inl ||| (inl.inr ||| inr)
   rassoc = (inl ||| inr.inl) ||| inr.inr
-instance (Cartesian co (ViaCocartesian co k), Cocartesian co k) => Symmetric co (ViaCocartesian co k) where
+instance Cocartesian co k => Symmetric co (ViaCocartesian co k) where
   swap = inr ||| inl
 
 -------------------------------------------------------------------------------
@@ -251,7 +250,7 @@ fork fs = rmap fs . dups
 unfork :: (CartesianR r p k, Obj2 k a b) => a `k` (p r b) -> r (a `k` b)
 unfork f = (. f) <$> exs
 
-pattern Fork :: (Obj' k f, Obj' k g, CartesianR h (:.:) k) => h (k f g) -> k f (h :.: g)
+pattern Fork :: (CartesianR h p k, Obj2 k f g) => h (k f g) -> k f (p h g)
 pattern Fork ms <- (unfork -> ms) where Fork = fork
 -- {-# complete Fork #-} -- See (:&) above
 
@@ -268,7 +267,7 @@ join fs = jams . rmap fs
 unjoin :: (CocartesianR r co k, Obj2 k a b) => co r a `k` b -> r (a `k` b)
 unjoin f = (f .) <$> ins
 
-pattern Join :: (Obj' k f, Obj' k g, CocartesianR h (:.:) k) => h (k f g) -> k (h :.: f) g
+pattern Join :: (CocartesianR h p k, Obj2 k f g) => h (k f g) -> k (p h f) g
 pattern Join ms <- (unjoin -> ms) where Join = join
 -- {-# complete Join #-} -- See (:&) above
 
@@ -319,10 +318,10 @@ instance MonoidalClosed (:*) (->) (->) where
   curry   = P.curry
   uncurry = P.uncurry
 
-deriving via (ViaCartesian   (:*) (->)) instance (Cocartesian (,) (ViaCartesian (,) (->))) => Associative (:*) (->)
-deriving via (ViaCartesian   (:*) (->)) instance (Cocartesian (,) (ViaCartesian (,) (->))) => Symmetric   (:*) (->)
-deriving via (ViaCocartesian (:+) (->)) instance (Cartesian Either (ViaCocartesian Either (->))) => Associative (:+) (->)
-deriving via (ViaCocartesian (:+) (->)) instance (Cartesian Either (ViaCocartesian Either (->))) => Symmetric   (:+) (->)
+deriving via (ViaCartesian   (:*) (->)) instance Associative (:*) (->)
+deriving via (ViaCartesian   (:*) (->)) instance Symmetric   (:*) (->)
+deriving via (ViaCocartesian (:+) (->)) instance Associative (:+) (->)
+deriving via (ViaCocartesian (:+) (->)) instance Symmetric   (:+) (->)
 
 instance Representable r => MonoidalR r Ap (->) where
   rmap rab (Ap ra) = Ap (liftR2 ($) rab ra)
--- a/src/Category/Isomorphism.hs
+++ b/src/Category/Isomorphism.hs
@@ -70,6 +70,17 @@ instance Closed e k => Closed e (Iso k) where
 -- | Utilities
 -------------------------------------------------------------------------------
 
+isoSwap :: Iso k a b -> Iso k b a
+isoSwap (f :<-> g) = g :<-> f
+
+-- | Apply one isomorphism via fwd
+apFwdIso :: a <-> b -> a -> b
+apFwdIso (f :<-> _) a = f a
+
+-- | Apply one isomorphism via rev
+apRevIso :: a <-> b -> b -> a
+apRevIso = apFwdIso . isoSwap
+
 -- | Apply one isomorphism via another
 via :: (Category k, Obj2 k a b) => Iso k b b -> Iso k a b -> Iso k a a
 (g :<-> g') `via` (ab :<-> ba) = ba . g . ab :<-> ba . g' . ab
--- a/src/Category/Isomorphism.hs
+++ b/src/Category/Isomorphism.hs
@@ -70,16 +70,8 @@ instance Closed e k => Closed e (Iso k) where
 -- | Utilities
 -------------------------------------------------------------------------------
 
-isoSwap :: Iso k a b -> Iso k b a
-isoSwap (f :<-> g) = g :<-> f
-
--- | Apply one isomorphism via fwd
-apFwdIso :: a <-> b -> a -> b
-apFwdIso (f :<-> _) a = f a
-
--- | Apply one isomorphism via rev
-apRevIso :: a <-> b -> b -> a
-apRevIso = apFwdIso . isoSwap
+inverse :: Iso k a b -> Iso k b a
+inverse (f :<-> g) = g :<-> f
 
 -- | Apply one isomorphism via another
 via :: (Category k, Obj2 k a b) => Iso k b b -> Iso k a b -> Iso k a a
--- a/src/Category/Isomorphism.hs
+++ b/src/Category/Isomorphism.hs
@@ -70,9 +70,6 @@ instance Closed e k => Closed e (Iso k) where
 -- | Utilities
 -------------------------------------------------------------------------------
 
-inverse :: Iso k a b -> Iso k b a
-inverse (f :<-> g) = g :<-> f
-
 -- | Apply one isomorphism via another
 via :: (Category k, Obj2 k a b) => Iso k b b -> Iso k a b -> Iso k a a
 (g :<-> g') `via` (ab :<-> ba) = ba . g . ab :<-> ba . g' . ab
--- a/src/InductiveMatrix.hs
+++ b/src/InductiveMatrix.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE UndecidableInstances #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
 {- |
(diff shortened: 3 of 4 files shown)


src/Category/Isomorphism.hs
```diff
@@ -70,6 +70,17 @@ instance Closed e k => Closed e (Iso k) where
 -- | Utilities
 -------------------------------------------------------------------------------
 
+isoSwap :: Iso k a b -> Iso k b a
+isoSwap (f :<-> g) = g :<-> f
```
conal, line 74: The name "swap" sounds like an operation on representation rather than on meaning.

conal: See `inv` ("inverse"). Maybe "`inverse`" is a better name choice.

src/Category/Isomorphism.hs
```diff
@@ -70,6 +70,17 @@ instance Closed e k => Closed e (Iso k) where
...
+
+-- | Apply one isomorphism via fwd
+apFwdIso :: a <-> b -> a -> b
+apFwdIso (f :<-> _) a = f a
+
+-- | Apply one isomorphism via rev
+apRevIso :: a <-> b -> b -> a
+apRevIso = apFwdIso . isoSwap
```
conal, line 82: See `isoFwd` and `isoRev` (implemented as selectors).

src/InductiveMatrix.hs
```diff
@@ -62,8 +63,85 @@ instance LinearMap L where
...
 
-instance Category (L s) where
+pattern Join :: (Additive s,
+                Foldable h,
+                Eq (Rep h),
+                C3 V f g h,
+                CocartesianR h (:.:) (L s))
+             => h (L s f g) -> L s (h :.: f) g
+pattern Join ms <- (unjoin -> ms) where Join = JoinL
+{-# complete Join #-}
```
conal, line 77: Better moved to `Category` (as with `(:|)` and `(:&)` now defined there).

conal, line 77: Better moved to `Category` as with `(:|)` and `(:&)` now defined there.

src/Category/Isomorphism.hs
bolt12: Yes, `inverse` is a lot better!

src/Category/Isomorphism.hs
bolt12: Thanks for the reference!

src/InductiveMatrix.hs
bolt12: I'll move to there!

bolt12:
Forgot to mention that I had to add `UndecidableInstances`:

See here why, for example:
```haskell
[bios] [E] • The constraint ‘CartesianR r (:.:) (L s)’
    is no smaller than the instance head ‘Additive (L s f g)’
  (Use UndecidableInstances to permit this)
• In the instance declaration for ‘Additive (L s f g)’
```

src/Category/Isomorphism.hs
```diff
@@ -70,6 +70,9 @@ instance Closed e k => Closed e (Iso k) where
 -- | Utilities
 -------------------------------------------------------------------------------
 
+inverse :: Iso k a b -> Iso k b a
+inverse (f :<-> g) = g :<-> f
```
conal, line 74: This operation is already there, as `inv` (which could be renamed to "`inverse`").

src/InductiveMatrix.hs
```diff
@@ -63,7 +64,74 @@ instance LinearMap L where
...
+          forall r. CocartesianR r (:.:) (L s),
+          Cartesian (:*:) (L s),
+          Cocartesian (:*:) (L s))
+          => Additive (L s f g) where
+  zero                 = isoRev rowMajIso . isoFwd rowMajIso $ zero
+  Scale s + Scale s'   = Scale (s + s')
+  (f :|# g) + (h :| k) = (f + h) :| (g + k)
+  (f :&# g) + (h :& k) = (f + h) :& (g + k)
+  (f :|# g) + m        = let (h, k) = unjoin2 m in (f + h :|# g + k)
+  (f :&# g) + m        = let (h, k) = unfork2 m in (f + h :&# g + k)
```
conal, line 77: These two lines are subsumed by the previous two, aren't they?

src/InductiveMatrix.hs
```diff
@@ -63,7 +64,74 @@ instance LinearMap L where
...
+  JoinL ms  + Join ms' = Join (ms +^ ms')
+  ForkL ms  + Fork ms' = Fork (ms +^ ms')
+  JoinL ms  + m = let ms' = unjoin m in Join (ms +^ ms')
+  ForkL ms  + m = let ms' = unfork m in Fork (ms +^ ms')
```
conal, line 81: See previous comment.

src/InductiveMatrix.hs
```diff
@@ -63,7 +64,74 @@ instance LinearMap L where
...
+
+rowMajIso :: Iso (->) (L s a b) (b (a s))
```
conal, line 83: Do you have uses in mind for this isomorphism?

If we do want this isomorphism, I would first make a new module for row-major non-inductive matrices that implements the shared vocabulary and then use a more general isomorphism between _any_ two linear map representations.

src/InductiveMatrix.hs
```diff
@@ -63,7 +64,74 @@ instance LinearMap L where
...
+    fwd (f :&# g) = fwd f :*: fwd g
+    fwd (JoinL m) = cotraverse Comp1 $ fmapRep fwd m
+    fwd (ForkL m) = Comp1 $ fmapRep fwd m
+    rev :: b (a s) -> L s a b
+    rev = undefined -- TODO: Write rev for toRowMajIso.
+
+instance (forall r. CartesianR r (:.:) (L s),
+          forall r. CocartesianR r (:.:) (L s),
+          Cartesian (:*:) (L s),
+          Cocartesian (:*:) (L s)
```
conal, line 98: These four constraints all look wrong to me.

src/InductiveMatrix.hs
```diff
@@ -63,7 +64,74 @@ instance LinearMap L where
...
+         ) => Category (L s) where
   type Obj' (L s) a = V a
-  id = undefined
-  (.) = undefined
+  id = undefined                                  -- TODO: Write id
+  Scale a   . Scale b   = Scale (a * b)           -- Scale denotation
+  (p :&# q) . m         = (p . m) :&# (q . m)     -- binary product law
+  m         . (p :|# q) = (m . p) :|# (m . q)     -- binary coproduct law
+  (r :|# s) . (p :&# q) = (r . p) + (s . q)       -- biproduct law
```
conal, line 105: I'd drop parens on the RHS, relying on readers knowing that `.` bind _very_ tightly.

src/InductiveMatrix.hs
```diff
@@ -63,7 +64,74 @@ instance LinearMap L where
...
+  ForkL ms' . m         = ForkL (fmap (. m) ms')  -- n-ary product law
+  m'        . JoinL ms  = JoinL (fmap (m' .) ms)  -- n-ary coproduct law
+  JoinL ms' . ForkL ms  = sum (liftR2 (.) ms' ms)   -- biproduct law
+
+instance (forall r. Representable r, Additive s)
+      => CartesianR r (:.:) (L s) where
+  exs = unfork id
```
conal, line 112: Take care in case you want to define `id = fork exs` or `id = join ins`.

conal, line 112: Take care in case you want to define `id = fork exs` or `id = join ins`.

src/InductiveMatrix.hs
```diff
@@ -63,7 +64,74 @@ instance LinearMap L where
...
+  dups = undefined -- TODO: Write dups
+
+instance (forall r. Representable r, Additive s)
+      => CocartesianR r (:.:) (L s) where
+  ins  = unjoin id
```
conal, line 117: Take care in case you want to define`id = join ins`.

src/InductiveMatrix.hs
```diff
@@ -63,7 +64,74 @@ instance LinearMap L where
...
+  jams = undefined -- TODO: write jams
+
+instance (forall r. Representable r, Additive s) => Monoidal (:*:) (L s) where
+  f ### g = (inl . f) :|# (inr . g)
```
conal, line 121: This definition or its dual (via `Cartesian` instead of `Cocartesian`) should probably be a default in `Category`.

src/InductiveMatrix.hs
conal: On second thought, I guess this `id` definition is not general enough to type-check.

src/InductiveMatrix.hs
conal: On second thought, I guess this `id` definition is not general enough to type-check (as with `id` via `exs`).

src/InductiveMatrix.hs
```diff
@@ -63,7 +64,74 @@ instance LinearMap L where
...
+
+instance (forall r. Representable r, Additive s) => Cartesian (:*:) (L s) where
+  exl = id :| zero
+  exr = zero :| id
+  dup = id :&# id
+
+instance (forall r. Representable r, Additive s) => Cocartesian (:*:) (L s) where
+  inl = id :&# zero
+  inr = zero :& id
+  jam = id :|# id
```
conal, line 134: Are definitions all laws for abelian categories? If so, they seem like good defaults to go into `Category`.

src/InductiveMatrix.hs
conal: If `b (a s)` were wrapped into a category (plus cartesian etc), then this `rowMajIso` would have a pretty, compelling, and clearly correct implementation.

src/InductiveMatrix.hs
conal: Since there are two natural defaults—as with `Symmetric` and `Associative`—@dwincort's `deriving via` technique applies.

conal:
I renamed this PR to fit its intent better and to make it easier to find later.

src/InductiveMatrix.hs
```diff
@@ -63,7 +64,74 @@ instance LinearMap L where
...
+  ForkL ms  + m = let ms' = unfork m in Fork (ms +^ ms')
+
+rowMajIso :: Iso (->) (L s a b) (b (a s))
+rowMajIso = fwd :<-> rev
+  where
+    fwd :: L s a b -> b (a s)
+    fwd (Scale s) = pureRep (pureRep s)
+    fwd (f :|# g) = liftR2 (:*:) (fwd f) (fwd g)
+    fwd (f :&# g) = fwd f :*: fwd g
+    fwd (JoinL m) = cotraverse Comp1 $ fmapRep fwd m
```
conal, line 90: I don't think there's ever a good reason to use `fmapRep` instead of `fmap`, except in *defining* `fmap`, but there are at least two good reasons to use `fmap` instead of `fmapRep`.

src/InductiveMatrix.hs
```diff
@@ -63,7 +64,74 @@ instance LinearMap L where
...
+    fwd (ForkL m) = Comp1 $ fmapRep fwd m
+    rev :: b (a s) -> L s a b
+    rev = undefined -- TODO: Write rev for toRowMajIso.
+
+instance (forall r. CartesianR r (:.:) (L s),
+          forall r. CocartesianR r (:.:) (L s),
+          Cartesian (:*:) (L s),
+          Cocartesian (:*:) (L s)
+         ) => Category (L s) where
```
conal, line 99: Is there a specification that this instance should fulfill?

src/InductiveMatrix.hs
```diff
@@ -63,7 +64,74 @@ instance LinearMap L where
...
+  ForkL ms' . m         = ForkL (fmap (. m) ms')  -- n-ary product law
+  m'        . JoinL ms  = JoinL (fmap (m' .) ms)  -- n-ary coproduct law
+  JoinL ms' . ForkL ms  = sum (liftR2 (.) ms' ms)   -- biproduct law
+
+instance (forall r. Representable r, Additive s)
+      => CartesianR r (:.:) (L s) where
+  exs = unfork id
+  dups = undefined -- TODO: Write dups
+
+instance (forall r. Representable r, Additive s)
```
conal, line 115: `forall r. Representable r` ?

I don't think this constraint could possibly be satisfied. How did you come up with it?

conal:
> Forgot to mention that I had to add `UndecidableInstances`:
> 
> 
> 
> See here why, for example:
> 
> ```haskell
> 
> [bios] [E] • The constraint ‘CartesianR r (:.:) (L s)’
> 
>     is no smaller than the instance head ‘Additive (L s f g)’
> 
>   (Use UndecidableInstances to permit this)
> 
> • In the instance declaration for ‘Additive (L s f g)’
> 
> ```

I bet there is a simpler constraint you can use instead and then eliminate that pragma. Moreover, GHC will probably tell you what that simpler constraint is if you remove the one you’ve given.

src/InductiveMatrix.hs
bolt12: I thought so too but GHC says that those two pattern matchings are missing, as well as the `JoinL` and `ForkL` counterparts so I added them.

src/Category/Isomorphism.hs
bolt12: Oh sorry for the misunderstanding. I removed it!

src/InductiveMatrix.hs
bolt12: Well this iso was the only way I could think about implementing certain things. I agree that this should go for a different module but for ad-hoc purposes I left it here to see how it would go while implementing!

src/InductiveMatrix.hs
bolt12: Why is that? These constraints were all asked by GHC :/

src/InductiveMatrix.hs
bolt12: got it!

src/InductiveMatrix.hs
bolt12: Oh I see! Cool, I'll add those to the `Category` module!

src/InductiveMatrix.hs
bolt12: I see! I'll change!

src/InductiveMatrix.hs
bolt12: How come? Shouldn't the category laws be fulfilled?

src/InductiveMatrix.hs
bolt12: GHC asked for it :(

bolt12:
Regarding this, GHC asked for it, which is weird! Maybe you can inspect it and see what's going on, I just plugged in the constraints that made it possible for the code to compile. I'll try and refactor using your suggestions and see if I can simplify it!

src/InductiveMatrix.hs
bolt12: It arises from the use of `(:|)` and `(:&)` I think.

src/InductiveMatrix.hs
bolt12: Nvm! You were right!

src/InductiveMatrix.hs
bolt12: I managed to cut this constraints too.

src/InductiveMatrix.hs
bolt12: I think so: https://en.wikipedia.org/wiki/Abelian_category#Definitions

bolt12:
Made a few corrections, I think the most critical point is the `Additive` instance which is needed for `(.)` and makes all the type constraints propagate.

src/InductiveMatrix.hs
conal: Are you asking whether the category laws are fulfilled by your instance? Or whether fulfilling the category laws is all the specification we need? Or maybe something else?

bolt12: I think this one:

> Or whether fulfilling the category laws is all the specification we need?

But I'm intrigued by why you asked this question!

conal: Fulfilling the category laws would only tell us that we have a category, not that we have the category we want, so no, those laws do not suffice.

We are claiming that inductive matrices represent linear functions/maps/transformations, i.e., we’re drawing an *analogy* between a data type and a mathematical ideal, which includes the language (vocabulary) shared by the two. The specification is that this analogy is precisely true, which is to say that the homomorphism property for every *operation* is true.

If the homomorphic specification holds, then *all* valid laws about linear maps expressible via the shared vocabulary will necessarily hold about inductive matrices as well, thanks to (a) denotational equality and (b) the nature of homomorphism. I encourage you to try proving this claim for at least two operations, e.g., `id` and `(.)`.

In other words, if we take care of the homomorphic specification, the laws will take care of themselves (“for free”, or more accurately “already paid for”). If we don’t take care of the homomorphic specification, then the laws won’t do it for us.

src/InductiveMatrix.hs
conal: > Why is that? These constraints were all asked by GHC :/

In what way did GHC ask for these constraints? What were the error messages?

src/InductiveMatrix.hs
conal: > GHC asked for it :( 

Hm. How did GHC ask?

src/Category.hs
```diff
@@ -198,13 +198,14 @@ instance Category k => Category (ViaCartesian p k) where
...
   id = ViaCartesian id
   ViaCartesian g . ViaCartesian f = ViaCartesian (g . f)
-deriving instance Monoidal  p k => Monoidal  p (ViaCartesian p k)
-deriving instance Cartesian p k => Cartesian p (ViaCartesian p k)
+instance (Category k, Cocartesian p (ViaCartesian p k)) => Monoidal p (ViaCartesian p k) where
+  f ### g = (inl . f) ||| (inr . g)
+deriving instance (Cartesian p k, Cocartesian p (ViaCartesian p k)) => Cartesian p (ViaCartesian p k)
 
-instance Cartesian p k => Associative p (ViaCartesian p k) where
+instance (Cartesian p k, Cocartesian p (ViaCartesian p k)) => Associative p (ViaCartesian p k) where
```
conal, line 205: Why `Cocartesian` here?

src/Category.hs
```diff
@@ -198,13 +198,14 @@ instance Category k => Category (ViaCartesian p k) where
...
   lassoc = second exl &&& (exr . exr)
   rassoc = (exl . exl) &&& first  exr
-instance Cartesian p k => Symmetric p (ViaCartesian p k) where
+instance (Cartesian p k, Cocartesian p (ViaCartesian p k)) => Symmetric p (ViaCartesian p k) where
```
conal, line 208: `Cocartesian`?

src/Category.hs
```diff
@@ -214,16 +215,16 @@ instance Category k => Category (ViaCocartesian p k) where
...
+  f ### g = (f . exl) &&& (g . exr)
+deriving instance (Cartesian co (ViaCocartesian co k), Cocartesian co k) => Cocartesian co (ViaCocartesian co k)
 
-instance Cocartesian co k => Associative co (ViaCocartesian co k) where
+instance (Cartesian co (ViaCocartesian co k), Cocartesian co k) => Associative co (ViaCocartesian co k) where
   lassoc = inl.inl ||| (inl.inr ||| inr)
   rassoc = (inl ||| inr.inl) ||| inr.inr
-instance Cocartesian co k => Symmetric co (ViaCocartesian co k) where
+instance (Cartesian co (ViaCocartesian co k), Cocartesian co k) => Symmetric co (ViaCocartesian co k) where
   swap = inr ||| inl
```
conal, line 226: Why `Cartesian` here? (Same question as above.)

src/Category.hs
```diff
@@ -318,10 +319,10 @@ instance MonoidalClosed (:*) (->) (->) where
...
   uncurry = P.uncurry
 
-deriving via (ViaCartesian   (:*) (->)) instance Associative (:*) (->)
-deriving via (ViaCartesian   (:*) (->)) instance Symmetric   (:*) (->)
-deriving via (ViaCocartesian (:+) (->)) instance Associative (:+) (->)
-deriving via (ViaCocartesian (:+) (->)) instance Symmetric   (:+) (->)
+deriving via (ViaCartesian   (:*) (->)) instance (Cocartesian (,) (ViaCartesian (,) (->))) => Associative (:*) (->)
+deriving via (ViaCartesian   (:*) (->)) instance (Cocartesian (,) (ViaCartesian (,) (->))) => Symmetric   (:*) (->)
+deriving via (ViaCocartesian (:+) (->)) instance (Cartesian Either (ViaCocartesian Either (->))) => Associative (:+) (->)
+deriving via (ViaCocartesian (:+) (->)) instance (Cartesian Either (ViaCocartesian Either (->))) => Symmetric   (:+) (->)
```
conal, line 325: Same question as above about the new constraints.

src/InductiveMatrix.hs
bolt12: Well the erro rmessages were something like `Couldn't infer <constraint` so I just added it. The problem is the usage of the pattern synonyms, I think.

src/InductiveMatrix.hs
bolt12: It gives a compile error saying that it couldn't infer `Representable r` and when I added it it said it couldn't infer `Representable r0` so I just plucked a `forall r` and GHC was happy. It still is very strange...

src/Category.hs
bolt12: I added an instance of `Monoidal` to `ViaCartesian` and `ViaCocartesian` just like you suggested and then the constraints just propagated to those places. I just plucked whatever GHC asked for and it compiled!

Nonetheless it felt very weird to do it. I think there is a certain limitation on the degree of polymorphism we can use reliably in this case.

src/InductiveMatrix.hs
conal: Got it; thanks. I’m afraid that practice (adding whatever constraints GHC couldn’t infer) results in hiding symptoms rather than solving underlying issues. Instead, I suggest asking yourself why *you* believe the missing constraints should hold. As you try to clarify your argument, you will hopefully find and fix the flaw in it. (See [*Literate Programming and the “Spaniel” Method*](https://asyropoulos.eu/papers/PDF/p52-hatzigeorgiu.pdf). More [here](https://www.skmurphy.com/blog/2015/09/25/asking-questions-from-a-caring-perspective/).) I’m happy to help you through this reasoning process.

src/InductiveMatrix.hs
conal: Although GHC is now happy with the definition, it will be unhappy with *all potential uses* of that definition, because those uses will have to prove that `forall r. Representable r`, which is false. To fix those uses, you would have to add this same unsatisfiable constraint, which breaks *their* uses, etc.

In term of logic, you can turn *any* proposition into a provable one by prepending `false =>`, but this altered proposition will not be applicable to any situation.

src/Category.hs
conal: Thanks. Let's try again without these adding these weird constraints.

bolt12: I just realized that I blundered and swapped the cartesian with cocartesian instances, that's why the constrains all look wrong. My bad. I'll fix it! Sorry for the trouble!

src/InductiveMatrix.hs
bolt12: I understand. One way would be to define this in terms of `mu` but I don't have the inverse. While trying to write code for the `InductiveMatrix` module I'm finding more and more that the "n-ary" counterparts of `join` and `fork` are a lot of the times the cause of obstacles regarding type inference and constraints. I'm not sure if this is a good sign too.

src/Category.hs
bolt12: Fixed!

src/Category.hs
bolt12: Fixed!

src/Category.hs
bolt12: Fixed!

src/Category.hs
bolt12: Also fixed with the latest push.

bolt12:
The bottleneck is the `Additive s` type class. I don't know of a way to write it without enabling `UndecidableInstances` and without making weird constraints propagate to the other instances.

src/InductiveMatrix.hs
conal: For clarity and succinctness, I'd keep the two lines you deleted (using pattern synonyms) and delete the two you kept. You'll get a completeness/exhaustiveness warning, which can be fixed by  adding `COMPLETE` pragmas as mentioned in another comment.

src/InductiveMatrix.hs
conal: And replace this instance definition with a `deriving` declaration, right? I'd make sure both `deriving` choices (via cartesian and cocartesian) compile without warning, and then comment out one of them with a comment that either works.

src/Category.hs
conal: Looks better. Did it type-check the other way? If so, something is fishy.

conal: There already is a `Cartesian p (ViaCartesian p k)` instance, so you can add the just preconditions for *that* instance instead, and then eliminate redundant constraints. Check out @dwincort's `Associative` and `Symmetric` instances.

src/Category.hs
```diff
@@ -198,14 +197,14 @@ instance Category k => Category (ViaCartesian p k) where
   type Obj' (ViaCartesian p k) a = Obj k a
   id = ViaCartesian id
   ViaCartesian g . ViaCartesian f = ViaCartesian (g . f)
-instance (Category k, Cocartesian p (ViaCartesian p k)) => Monoidal p (ViaCartesian p k) where
-  f ### g = (inl . f) ||| (inr . g)
-deriving instance (Cartesian p k, Cocartesian p (ViaCartesian p k)) => Cartesian p (ViaCartesian p k)
+instance (Category k, Cartesian p (ViaCartesian p k)) => Monoidal p (ViaCartesian p k) where
```
conal, line 200: There already is a `Cartesian p (ViaCartesian p k)` instance, so you can add the just preconditions for *that* instance instead, and then eliminate redundant constraints. Check out @dwincort's `Associative` and `Symmetric` instances.

src/Category.hs
```diff
@@ -215,14 +214,14 @@ instance Category k => Category (ViaCocartesian p k) where
   type Obj' (ViaCocartesian p k) a = Obj k a
   id = ViaCocartesian id
   ViaCocartesian g . ViaCocartesian f = ViaCocartesian (g . f)
-instance (Category k, Cartesian co (ViaCocartesian co k)) => Monoidal co (ViaCocartesian co k) where
-  f ### g = (f . exl) &&& (g . exr)
-deriving instance (Cartesian co (ViaCocartesian co k), Cocartesian co k) => Cocartesian co (ViaCocartesian co k)
+instance (Category k, Cocartesian co (ViaCocartesian co k)) => Monoidal co (ViaCocartesian co k) where
```
conal, line 217: See comment about `Cartesian p (ViaCartesian p k)`.

src/Category.hs
```diff
@@ -251,7 +250,7 @@ fork fs = rmap fs . dups
 unfork :: (CartesianR r p k, Obj2 k a b) => a `k` (p r b) -> r (a `k` b)
 unfork f = (. f) <$> exs
 
-pattern Fork :: (Obj' k f, Obj' k g, CartesianR h (:.:) k) => h (k f g) -> k f (h :.: g)
+pattern Fork :: (CartesianR h p k, Obj2 k f g) => h (k f g) -> k f (p h g)
```
conal, line 253: Good catch.

src/InductiveMatrix.hs
```diff
@@ -112,39 +112,39 @@ instance () => Category (L s) where
...
 -- TODO: Derive Via
-instance (forall r. Representable r, Additive s) => Monoidal (:*:) (L s) where
+instance Additive s => Monoidal (:*:) (L s) where
   f ### g = (inl . f) :|# (inr . g)
 
 -- TODO: Add deriving via capabilities.
 -- Couldn't get constraints to work when defining this instance on the Via
 -- type in Category.hs
-instance (Representable r, Cartesian (:.:) (L s), Additive s) => MonoidalR r (:.:) (L s) where
+instance (Representable r, Cartesian (:.:) (L s)) => MonoidalR r (:.:) (L s) where
```
conal, line 132: You can simply `Cartesian (:.:) (L s)` here.

conal: If you remove the constraint in this case, I think GHC will suggest the simplest replacement.

src/InductiveMatrix.hs
```diff
@@ -112,39 +112,39 @@ instance () => Category (L s) where
...
 instance (Representable r, Cartesian (:.:) (L s), Additive s)
       => CocartesianR r (:.:) (L s) where
   ins  = unjoin id
-  jams = undefined -- TODO: write jams
+  jams = undefined -- TODO: Write jams
 
 -- TODO: Derive Via
-instance (forall r. Representable r, Additive s) => Monoidal (:*:) (L s) where
+instance Additive s => Monoidal (:*:) (L s) where
   f ### g = (inl . f) :|# (inr . g)
```
conal: Now you can replace with `deriving via` here in two different ways, right? I'd make sure both type-check, and then comment one out, prefixed by "`-- Equivalently, `".

src/InductiveMatrix.hs
```diff
@@ -112,39 +112,39 @@ instance () => Category (L s) where
   m'        . JoinL ms  = JoinL (fmap (m' .) ms)  -- n-ary coproduct law
   JoinL ms' . ForkL ms  = sum (liftR2 (.) ms' ms)   -- biproduct law
 
-instance (Representable r,Cartesian (:.:) (L s) , Additive s)
+instance (Representable r, Cartesian (:.:) (L s) , Additive s)
```
conal, line 115: You can simplify `Cartesian (:.:) (L s)`, since there's an instance defined.

src/InductiveMatrix.hs
```diff
@@ -112,39 +112,39 @@ instance () => Category (L s) where
...
       => CartesianR r (:.:) (L s) where
   exs = unfork id
   dups = undefined -- TODO: Write dups
 
 instance (Representable r, Cartesian (:.:) (L s), Additive s)
```
conal, line 120: You can simplify `Cartesian (:.:) (L s)`, since there's an instance defined.

src/InductiveMatrix.hs
conal: Now I see the `TODO`.

src/InductiveMatrix.hs
```diff
@@ -112,39 +112,39 @@ instance () => Category (L s) where
...
   exr = zero :| id
   dup = id :&# id
 
 -- TODO: Move to Category.hs
 -- See: https://en.wikipedia.org/wiki/Abelian_category#Definitions
-instance (forall r. Representable r, Additive s) => Cocartesian (:*:) (L s) where
+instance Additive s => Cocartesian (:*:) (L s) where
   inl = id :&# zero
   inr = zero :& id
   jam = id :|# id
```
conal: More potential defaults for `Category`.

src/InductiveMatrix.hs
conal: Something else is wrong here, since `Cartesian` goes with `(:*:)`, and `CartesianR` goes with `(:.:)` but the error message is about `Cartesian (:.:) (L s)`.

conal: Even weirder, the error message I see says "Could not deduce (Cartesian (:.:) (L s)) arising from a use of ‘unjoin’". And I'd expect `unjoin` to be about `CocartesianR`.

src/InductiveMatrix.hs
bolt12: I'm not very familiar with how pattern synonyms work, how can I give a typed COMPLETE pragma with each cartesian category instance?

src/InductiveMatrix.hs
bolt12: Yes I see the same error! It is super weird! I don't know where that comes from.

src/Category.hs
bolt12: I see! Thanks!

src/Category.hs
bolt12: Thanks :smile:

src/InductiveMatrix.hs
bolt12: It's weird because, to me, GHC asks for `Cartesian (:.:) (L s)`, is it not the case with you? How can I simplify it?

conal: Now I notice the unexpected combination of `Cartesian` and `(:.:)`. Looking at this `MonoidalR` instance, I asked myself how `Cartesian` gets involved. Aha! The `rmap` definition uses `exr`.

bolt12: But looking at it now.. does `Cartesian (:.:) (L s)` make sense/exist? I need to give an instance for it right?

conal: I suggest commenting (or `ifdef`ing) out what code you can, for instance all of the n-ary instances, while getting the rest into good shape. Then reactivate one instance at a time.

src/InductiveMatrix.hs
bolt12: There's an instance for `Cartesian (:*:) (L s)`, not `(:.:)`, right? This also doesn't seem right!

src/InductiveMatrix.hs
conal: Once this bogus `Cartesian (:.:) (L s)` got into your `MonoidalR` instance, it infected your `CartesianR` and `CocartesianR` instances.

src/InductiveMatrix.hs
bolt12: This constraint is propagated by `MonoidalR`.

src/InductiveMatrix.hs
conal: > But looking at it now.. does `Cartesian (:.:) (L s)` make sense/exist? I need to give an instance for it right? 

It definitely does not make sense, so no, don't give an instance.

conal: I think you just need to fix your `rmap` definition and remove all of those bogus `Cartesian (:.:) (L s)` constraints.

src/InductiveMatrix.hs
conal: I agree. Eliminate that constraint rather than simplifying it.

src/InductiveMatrix.hs
bolt12: Okay! Thanks! I'll try it tomorrow! :smile:

conal: I guess one lesson here is to bring some skepticism to GHC's missing constraint messages. If the constraint doesn't seem to fit, then maybe the implementation is problematic. If so, change the implementation instead of adding constraints.

src/Category.hs
```diff
@@ -83,31 +83,25 @@ second :: (Monoidal p k, Obj3 k a b d) => (b `k` d) -> ((a `p` b) `k` (a `p` d))
 second g = id ### g
 
 class Monoidal p k => Cartesian p k where
+  {-# MINIMAL exl, exr #-}
```
conal, line 86: `(&&&)` *or* `dup` also needed.

conal: Likewise for `Cocartesian` and n-ary variants.

src/Category.hs
```diff
@@ -83,31 +83,25 @@ second :: (Monoidal p k, Obj3 k a b d) => (b `k` d) -> ((a `p` b) `k` (a `p` d))
...
-      => (a `k` c) -> (a `k` d) -> (a `k` (c `p` d))
-f &&& g = (f ### g) . dup
+  dup = id &&& id
+  infixr 3 &&&
+  -- Binary fork
+  (&&&) :: Obj3 k a c d => (a `k` c) -> (a `k` d) -> (a `k` (c `p` d))
+  f &&& g = (f ### g) . dup
+  -- Inverse of fork2
+  unfork2 :: Obj3 k a c d => (a `k` (c `p` d)) -> ((a `k` c) :* (a `k` d))
+  unfork2 f = (exl . f , exr . f)
```
conal, line 97: Also default definitions for `exl` and `exr` via `unfork2` and likewise for `Cocartesian` and the n-ary variants. And suitable `MINIMAL` declarations. At least that was the idea.

conal: Note the definitions of `ins` and `exs` in `LinAlg.hs`.

conal: I'm unsure about whether these default are really the ones we want for injections and projections. `LinAlg.hs` has a different style of definition for `inl`, `inr,` `exl`, and `exr`, which I think can also generalize to arbitrary abelian categories.

bgamari:
@conal, while I suspect it wouldn't be too technically challenging to relax the restriction, I do wonder about how usable the result would be. Afterall, the user would receive different warnings depending upon what they have imported; while users we grudgingly accept this out of necessity in the case of instances, it seems quite fiddly to allow for something that only throws a warning.

Is the reason for these pragams being orphans is the existence of import cycles? If so, is there anything preventing you from using the usual tools for breaking such cycles (e.g. `hs-boot` files and splitting up modules).

src/Category.hs
bolt12: Forgot those! Thanks!

src/Category.hs
bolt12: Regarding the Abelian Category I was thinking of using the same strategy as `ViaCartesian`/`ViaCocartesian` and have a `ViaAbelian` to derive everything for us. I think this approach would be the easier and require less work. What do you think?

bolt12:
@bgamari 

> Is the reason for these pragams being orphans is the existence of import cycles?

The primary reason of the pragmas being orphans is the fact that they are defined in a separate haskell module where it makes sense to define them (since they are polymorphic over a `Cartesian` and `Cocartesian` category we'd like to have these pattern synonyms for every type that instanciates this type classes). The problem as @conal  mentioned is that these pattern synonyms can't be made complete without being of a specific type, which when we try to do so, _in a different module_, GHC gives an error saying that we orphan pragmas are not supported.

We are not able to fix this without having cyclic imports (Since `Category.hs` has all the interface type classes and vocabulary, and `InductiveMatrix.hs` (which imports `Category.hs`), for example, has the specific type needed to make the pattern synonym complete).

> If so, is there anything preventing you from using the usual tools for breaking such cycles (e.g. hs-boot files and splitting up modules).

I'm not familiar with this strategy and given that we haven't tried it yet, it is worth giving it a shot. Could you expand on this or reference some kind of tutorial?

dwincort:
@bgamari I think another way to phrase what we're after is something like "associated pattern synonyms".  Let me describe a simple, minimal example:
```haskell
class Splittable a b c where
  split :: a -> (b, c)
  unsplit :: b -> c -> a

instance Splittable (b, c) b c where
  split = id
  unsplit = (,)

pattern Split :: Splittable a b c => b -> c -> a
pattern Split b c <- (split -> (b, c)) where Split = unsplit
{-# COMPLETE Split :: (,) #-}
```
Here, we have a type class, an instance of it, and a pattern keyed off of the class.  We cannot provide a general `COMPLETE` pragma for `Split`, for if we do, we get the GHC error "A type signature must be provided for a set of polymorphic pattern synonyms."  But, for `(,)`, which is an instance of `Splittable`, we _can_ provide a monomorphic complete pragma.

Now, say we wanted to package up `Splittable` and `Split` as a library for others to use.  The idea is that they can take their own data type, make it an instance of `Splittable` and then have access to the `Split` pattern.  Indeed, this works, but every use of `Split` will be plagued with incomplete pattern match warnings.  We cannot ask the user to place their complete pragma in the library code -- even if hs-boot files could work between libraries, it is still too much to ask that every user make a PR on the library just to get the functionality they need.

The next best solution we could think of (but one that doesn't work) was to ask users to put a `{-# COMPLETE Split :: T #-}` pragma right at their instance definition in their own code.  But, that doesn't work, yielding the orphan pragma error which is the basis of our question to you.  A more advanced option would definitely be something like "associated pattern synonyms" that I mentioned at the start of this post:
```haskell
class Splittable a b c where
  split :: a -> (b, c)
  unsplit :: b -> c -> a
  pattern Split :: Splittable a b c => b -> c -> a
  pattern Split b c <- (split -> (b, c)) where Split = unsplit
  {-# COMPLETE Split :: a #-}
```
That seems like a much bigger change, though hopefully it provides some intuition about what we're trying to do.

src/Category.hs
conal: Since there is more than one sensible default definition for injections and projections, I guess we should either (a) pick one, (b) do as with `ViaCartesian` etc, or (c) provide some definitions to be used in instances (as `Data.Functor.Rep` does).

bgamari:
> @bgamari
> 
> > Is the reason for these pragams being orphans is the existence of import cycles?
> 
> The primary reason of the pragmas being orphans is the fact that they are defined in a separate haskell module where it makes sense to define them (since they are polymorphic over a `Cartesian` and `Cocartesian` category we'd like to have these pattern synonyms for every type that instanciates this type classes). The problem as @conal mentioned is that these pattern synonyms can't be made complete without being of a specific type, which when we try to do so, _in a different module_, GHC gives an error saying that we orphan pragmas are not supported.
> 
> We are not able to fix this without having cyclic imports (Since `Category.hs` has all the interface type classes and vocabulary, and `InductiveMatrix.hs` (which imports `Category.hs`), for example, has the specific type needed to make the pattern synonym complete).
> 
> > If so, is there anything preventing you from using the usual tools for breaking such cycles (e.g. hs-boot files and splitting up modules).
> 
> I'm not familiar with this strategy and given that we haven't tried it yet, it is worth giving it a shot. Could you expand on this or reference some kind of tutorial?

See [Section 5.8 of the GHC users guide](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/separate_compilation.html#how-to-compile-mutually-recursive-modules).

bgamari:
> @bgamari I think another way to phrase what we're after is something like "associated pattern synonyms".

Indeed this is an idea that has come up before. See [GHC #8583](https://gitlab.haskell.org/ghc/ghc/-/issues/8583) for one such proposal. There are a few related ideas described in linked tickets.

dwincort:
@bgamari Yes, if we had those associated pattern synonyms, then (as long as they supported appropriate complete pragmas) we'd be good to go.  That said, associated pattern synonyms are perhaps _more_ than we need, since orphan complete pragmas would do the job well enough.  GHC already lets us write the patterns we want — the thing we're missing here is the ability to write the complete pragmas for those patterns.

src/Category.hs
conal: Oops. I guess I added the previous remark with having seen your reply from two hours earlier. Yes, `ViaAbelian` seems a good idea to me.

src/Category.hs
conal: Thanks. Likewise, I think `exl` and `exr` are alternatives to `unfork2`.

conal: and vice versa.

conal: It might be prettier to replace the `exl` and `exr` methods with `ex2`, with default `ex2 = unfork id`. Compare with `exs` and `unfork`.

src/Category.hs
bolt12: Okay! Nice 😄

src/Category.hs
conal: We can still define `(exl,exr) = ex2` as a utility (rather than a method).

src/InductiveMatrix.hs
```diff
@@ -146,17 +146,15 @@ instance Representable r => MonoidalR r (:.:) (L s) where
 -- TODO: Move to Category.hs
 -- See: https://en.wikipedia.org/wiki/Abelian_category#Definitions
 instance Additive s => Cartesian (:*:) (L s) where
```
conal, line 148: We're missing `dup` or `(&&&)`.

src/InductiveMatrix.hs
```diff
@@ -146,17 +146,15 @@ instance Representable r => MonoidalR r (:.:) (L s) where
...
-  exl = id :| zero
-  exr = zero :| id
-  dup = id :&# id
+  exl = id :|# zero
+  exr = zero :|# id
 -- {-# COMPLETE (:&) :: L #-} -- See complete pragma above
 
 -- TODO: Move to Category.hs
 -- See: https://en.wikipedia.org/wiki/Abelian_category#Definitions
 instance Additive s => Cocartesian (:*:) (L s) where
```
conal, line 155: We're missing `jam` or `(|||)`.

conal:
@bgamari Thanks very much for chiming in\!

> Is the reason for these pragams being orphans is the existence of import cycles? If so, is there anything preventing you from using the usual tools for breaking such cycles (e.g. `hs-boot` files and splitting up modules).

The fundamental problem I see with this solution is that it breaks modularity *across packages*.

The pattern synonym belongs to a quite general *class* (cartesian categories) provided by a library, while the particular concrete instances arise in instances naturally found outside of that library. Currently we are defining the class and its instances in the same library, so mutually recursive modules may be an option but only temporarily.

Breaking modularity across packages seems like a big deal to me. Consider that the pattern synonyms we’re talking about could go in `Control.Arrow` for the `Arrow` and `ArrowChoice` classes, and the `COMPLETE` pragmas in every `Arrow` and `ArrowChoice` instance in the distributed Haskell ecosystem, including experimental, unreleased instances.

conal:
one above be supported> While I suspect it wouldn’t be too technically challenging to relax the restriction, I do wonder about how usable the result would be. Afterall, the user would receive different warnings depending upon what they have imported; while users we grudgingly accept this out of necessity in the case of instances, it seems quite fiddly to allow for something that only throws a warning.

@bgamari: I don’t think I understand what you’re envisioning about warnings and imports.

Really, I don’t ever want *any* warnings for these pattern synonyms, because they’ll always be complete.

Here is one of the isomorphisms in question:

``` haskell
infixr 3 &&&
(&&&) :: (Cartesian p k, Obj3 k a c d)
      => (a `k` c) -> (a `k` d) -> (a `k` (c `p` d))     -- "fork2"

-- Inverse of fork2
unfork2 :: (Cartesian p k, Obj3 k a c d)
        => (a `k` (c `p` d)) -> ((a `k` c) :* (a `k` d))

pattern (:&) :: (Cartesian p k, Obj3 k a c d)
             => (a `k` c) -> (a `k` d) -> (a `k` (c `p` d))
pattern f :& g <- (unfork2 -> (f,g)) where (:&) = (&&&)
-- {-# COMPLETE (:&) #-}  -- GHC rejects :(
```

where `Cartesian p k` means that `k` is a cartesian category with `p` as its products (similar to `Arrow` but lacking `arr`, poly-kinded, and without `(,)` wired in for products). It’s part of the nature of these categories that `(&&&)` is invertible, and so `(:&)` makes for much more declarative/readable programming, sharing a notation for combining and separating functions. For this reason, we want the pattern `f :& g` to be treated as complete *whenever* matching against a product-producing arrow.

Hm. Now I realize that even if GHC allowed `COMPLETE (:&) :: K` pragmas for individual `Cartesian` instances `K` (almost all of which are declared outside of the module—and eventually the package—defining `(:&)`), GHC would still object to all `Cartesian`-polymorphic code (i.e., working on arbitrary cartesian categories), since the allowable `COMPLETE` declarations are all too monomorphic. Maybe this polymorphic use is what you were getting at?

Do you think GHC could be made smarter about pattern synonyms so that we no longer need `COMPLETE` pragmas in cases like `(:&)`, with reasonably little work? Alternatively, could GHC support polymorphic `COMPLETE` pragmas such as `(:&)` above?

src/Category.hs
bolt12: How would this be an improvement? Is it just to be similar to the `R` variant?

src/InductiveMatrix.hs
bolt12: Fixed!

src/InductiveMatrix.hs
bolt12: Fixed!

src/Category.hs
conal: > How would this be an improvement? Is it just to be similar to the `R` variant?

In part to match the `R` variant, but more to explore exactly the question you’re asking here. I suspect this change would then suggest pretty reformulations of other definitions and laws and perhaps new insights. Many (perhaps most) of my insights come from noticing opportunities to say things in a prettier (more poetic) way and playing with the consequences.

Please consider my remark as my thinking out loud rather than a suggestion for this PR.

src/InductiveMatrix.hs
```diff
@@ -148,13 +148,15 @@ instance Representable r => MonoidalR r (:.:) (L s) where
 instance Additive s => Cartesian (:*:) (L s) where
   exl = id :|# zero
   exr = zero :|# id
```
conal: Or define `unfork2`, and rely on the defaults for `exl` and `exr`.

src/InductiveMatrix.hs
```diff
@@ -148,13 +148,15 @@ instance Representable r => MonoidalR r (:.:) (L s) where
...
+  (&&&) = (:&#)
 -- {-# COMPLETE (:&) :: L #-} -- See complete pragma above
 
 -- TODO: Move to Category.hs
 -- See: https://en.wikipedia.org/wiki/Abelian_category#Definitions
 instance Additive s => Cocartesian (:*:) (L s) where
   inl = id :&# zero
   inr = zero :&# id
```
conal: Or define `unjoin2`, and rely on the defaults for `inl` and `inr`.

I guess the choice will depend on how we want to define `id`.

conal: We'll need to fix the `MINIMAL` declarations in the binary and n-ary cartesian and cocartesian classes and add some default definitions there.

src/Category.hs
bolt12: I see! Then I'd suggest that when/if this is able to be merged we open a new PR exploring this!

src/InductiveMatrix.hs
bolt12: > We'll need to fix the MINIMAL declarations in the binary and n-ary cartesian and cocartesian classes and add some default definitions there.

I'm not sure I understand this. Are you suggesting to change the MINIMAL declarations to something like this, for example?

```haskell
class Monoidal p k => Cartesian p k where
  {-# MINIMAL (exl, exr, ((&&&) | dup) | (unfork2, ((&&&) | dup))) #-}
```

conal: We can define (a) `unfork2` and `(exl,exr)` via each other, and (b) `(&&&)` and `dup` via each other, 
Equivalently to what you wrote, but more tersely,

```haskell
  {-# MINIMAL ((exl,exr) | unfork2), ((&&&) | dup) #-}

Comment (bolt12) [reply src/InductiveMatrix.hs]: Got it, I'm on it!

I am having trouble defining custom instances for the `R` variant, this is what I have so far:

```haskell
class MonoidalR r p k => CartesianR r p k where
  {-# MINIMAL ((exs | unfork), (fork | dups)) #-}
  exs  :: Obj k a => r (p r a `k` a)
  exs = askRep $ unfork id
  dups :: Obj k a => a `k` p r a
  dups = fork (_ id)
  fork :: Obj2 k a c => r (a `k` c) -> (a `k` p r c)
  fork fs = rmap fs . dups
  unfork :: Obj2 k a b => a `k` (p r b) -> r (a `k` b)
  unfork f = (. f) <$> exs

class MonoidalR r co k => CocartesianR r co k where
  {-# MINIMAL ((ins | unjoin), (join | jams)) #-}
  ins  :: Obj k a => r (a `k` co r a)
  ins = askRep $ unjoin id
  jams :: Obj k a => co r a `k` a
  jams = join (_ id)
  join :: Obj2 k a b => r (a `k` b) -> co r a `k` b
  join fs = jams . rmap fs
  unjoin :: Obj2 k a b => co r a `k` b -> r (a `k` b)
  unjoin f = (f .) <$> ins
```

The `dups` and `jams` are the ones I can not solve. The type holes do not help as they did with `ins` and `exs`. Any thoughts?

conal: Noting `dup = id &&& id`, I’d try `dups = fork (pureRep id)`. Similarly for `jam` and `jams`.

bolt12:
I tried it but it requires `Representable r` constraint!

src/InductiveMatrix.hs
conal: Since these default `dups` and `jams` definitions presume `Representable r`, which `MonoidalR r p k` does not have as a superclass constraint, we’d need to either add that superclass or add default signatures for `dups` and `jams`. Either way seems fine to me for now.

bolt12: Would a better approach be having:

```haskell
class    (Representable r, ObjR' r p k) => ObjR r p k
instance (Representable r, ObjR' r p k) => ObjR r p k
```

instead of

```haskell
class    (Functor r, ObjR' r p k) => ObjR r p k
instance (Functor r, ObjR' r p k) => ObjR r p k
```

conal: I like that approach even better!

bolt12: Awesome! Thanks!

bolt12: Now, I found another hurdle which is the following:

```haskell
instance (Representable r) => CocartesianR r (:.:) (L s) where
  ins  = unjoin id
  join = JoinL -- ERROR: Needs Eq (Rep r) and UndecidableInstances consequentially
```

Note that there still exists the circular dependency of `ins` and `unjoin`, for example.

conal: > 
> 
> ``` haskell
>   join = JoinL -- ERROR: Needs Eq (Rep r) and UndecidableInstances consequentially
> ```

It’s okay for us to `UndecidableInstances` to the module and add `Eq (Rep r)` to `Obj' (L s) r`.

> Note that there still exists the circular dependency of `ins` and `unjoin`, for example.

This sort of circularity is fine for defaults (and is why GHC Haskell has `MINIMAL` pragmas), but we’ll need to break the cycles for each instance.

conal: Since inductive matrices are defined around (binary and n-ary) fork & join, I assume we’d want to define `(&&&)`, `unfork2`, `(|||)`, and `unjoin2`.

bolt12: > It’s okay for us to UndecidableInstances to the module and add Eq (Rep r) to Obj' (L s) r

How would I do that? Isn't that restriction for `Category` class? 

> Since inductive matrices are defined around (binary and n-ary) fork & join, I assume we’d want to define (&&&), unfork2, (|||), and unjoin2, and use defaults for exl, exr, dup, inl, inr, and jam. Likewise for n-ary.

I'm having a hard time writing `unjoin` and `unfork`. Could you help me? I'll push what I have.

conal: > > It’s okay for us to add `UndecidableInstances` to the module and add `Eq (Rep r)` to `Obj' (L s) r`.
> 
> How would I do that? Isn’t that restriction for `Category` class?

Currently in `InductiveMatrix`:

``` haskell
type V = Representable

...
instance Category (L s) where
  type Obj' (L s) a = V a
  ...  
```

I’m suggesting the V definition as follows:

``` haskell
type V' a = (Representable a, Eq (Rep a))

class    V' a => V a
instance V' a => V a
```

Then change other `Representable` constraints to `V` as needed.

> I’m having a hard time writing `unjoin` and `unfork`. Could you help me? I’ll push what I have.

Happy to. Just let me know when you’ve pushed.

conal: > I’m having a hard time writing `unjoin` and `unfork`. Could you help me? I’ll push what I have.

We already have definitions `unjoinL` and `unforkL` in `LinAlg`.

bolt12: @conal I already pushed it!

bolt12: > We already have definitions unjoinL and unforkL (along with unjoin2 and unfork2) in LinAlg

Oh good catch! Sorry for the trouble!

bolt12: Regarding what you suggested, it doesn't work as expected, I tried it and GHC still asks to add the `Eq (Rep r)` constraint in the `R` type classes.

conal: How about replacing the `Representable r` superclass constraint by `V r` in those instances? For instance,

``` haskell
instance V r => CartesianR r (:.:) (L s) where
  ...
```

bolt12: Yes that simplifies things, nice suggestion! Thanks!

bolt12:
@conal I've pushed the latest changes, I think this is going well, the only major TODO is writing `rev` , the reverse isomorphism function of `rowMajIso`! Maybe you could go through the latest changes and see if everything looks good and then we can discuss the best way to write `rev`, what do you think? :)

src/InductiveMatrix.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE UndecidableInstances #-}
 {-# LANGUAGE ScopedTypeVariables #-}
```
conal, line 2: I'd activate (uncomment) `ScopedTypeVariables` in package.yaml.

src/InductiveMatrix.hs
```diff
@@ -65,23 +69,9 @@ instance LinearMap L where
...
---  is no smaller than the instance head ‘Additive (L s f g)’
---  (Use UndecidableInstances to permit this)
---
--- And then it will ask for Representable r, for which we need to give
--- a quantified constrain: forall r. Representable r.
-instance Additive s => Additive (L s f g) where
+-- This gives non-exhaustive pattern matching because pattern synonyms
+-- cannot be made complete
+instance Semiring s => Additive (L s f g) where
   zero                 = isoRev rowMajIso . isoFwd rowMajIso $ zero
```
conal, line 75: I think this definition is circular, with the RHS `zero` being at the same type as the LHS `zero`. To make sure, I replaced RHS `zero` by a hole. Also, there's a law that `isoRev iso . isoFwd iso == id`.

conal: See `zeroL` in `LinAlg`.

src/InductiveMatrix.hs
```diff
@@ -112,50 +106,54 @@ instance Category (L s) where
   m'        . JoinL ms  = JoinL (fmap (m' .) ms)  -- n-ary coproduct law
   JoinL ms' . ForkL ms  = sum (liftR2 (.) ms' ms) -- biproduct law
 
-instance Representable r => CartesianR r (:.:) (L s) where
+instance (Representable r, Eq (Rep r), Semiring s) => CartesianR r (:.:) (L s) where
```
conal: Could use `V r`.

src/InductiveMatrix.hs
```diff
@@ -112,50 +106,54 @@ instance Category (L s) where
...
+instance (Representable r, Eq (Rep r), Foldable r, Semiring s) => CocartesianR r (:.:) (L s) where
+  join = JoinL -- Needs Eq (Rep r) and UndecidableInstances consequentially
+  unjoin (p :&# p') = liftR2 (:&#) (unjoin p) (unjoin p')
+  unjoin (JoinL ms) = ms
+  unjoin (ForkL ms) = fmap ForkL (distribute (fmap unjoin ms))
 -- {-# COMPLETE Join :: L #-} -- See complete pragma above
 
--- TODO: Derive Via
-instance Additive s => Monoidal (:*:) (L s) where
+-- TODO: Add deriving capabilities
```
conal, line 124: Did you try `deriving via` here?

src/InductiveMatrix.hs
```diff
@@ -112,50 +106,54 @@ instance Category (L s) where
...
+instance Semiring s => Monoidal (:*:) (L s) where
   f ### g = (inl . f) :|# (inr . g)
 -- deriving via (ViaCartesian (:*:) (L s)) instance Monoidal (:*:) (L s)
 
 -- TODO: Add deriving via capabilities.
 -- Couldn't get constraints to work when defining this instance on the Via
 -- type in Category.hs
-instance Representable r => MonoidalR r (:.:) (L s) where
+instance (Representable r, Eq (Rep r), Semiring s) => MonoidalR r (:.:) (L s) where
```
conal, line 132: Could use `V r`.

src/InductiveMatrix.hs
```diff
@@ -112,50 +106,54 @@ instance Category (L s) where
...
 --   If that's it, then -ddump-splices might be useful
 --
 -- deriving via (ViaCartesian (:.:) (L s)) instance () => (MonoidalR r (:.:) (L s))
 
 -- TODO: Move to Category.hs
 -- See: https://en.wikipedia.org/wiki/Abelian_category#Definitions
-instance Additive s => Cartesian (:*:) (L s) where
+instance Semiring s => Cartesian (:*:) (L s) where
   (&&&) = (:&#)
   unfork2 f = ((id :|# zero) . f, (zero :|# id) . f)
```
conal, line 148: This `unfork2` definition is a hybrid of the defaults for `unfork2`, `exl`, and `exr`. I'd either define `unfork2` *or* `exl` and `exr` as in `LinAlg`.

conal: If the latter (`exl` and `exr`), then consider how the default `unfork2` plays out, especially with `(.)`. Beware cycles between `(.)` and `unfork2`.

src/InductiveMatrix.hs
conal: We might delay abbreviating `Representable r, Eq (Rep r)` to `V r` until we see whether `V` needs additional constraints, so as avoid unnecessarily restricting the inductive matrix category.

conal: Never mind. I see your second commit.

conal:
I think we can fix `id` and `rowMajIso` as in `LinAlg`.

There may also be a prettier way to get `rowMajIso` using a technique @dwincort suggested: define a GADT of run-time representations of representable functors and a class of representable functors that can generate this representation. Then generate `rowMajIso` and many other useful operations from these GADT values. I’d postpone this improvement to another PR and just steal from `LinAlg` for now.

conal:
I suggest postponing the missing `isoRev mu` (in the `LinearMap L` instance) for another PR also. I think there’s a definition to be found that works for *all* linear map representations.

src/InductiveMatrix.hs
bolt12: Yes, that's probably better, didn't even remembered that those definitions existed! Thanks!

src/InductiveMatrix.hs
bolt12: Yes, you're right. I don't know what I was thinking!

bolt12:
> There may also be a prettier way to get `rowMajIso` using a technique @dwincort suggested: define a GADT of run-time representations of representable functors and a class of representable functors that can generate this representation. Then generate `rowMajIso` and many other useful operations from these GADT values. I’d postpone this improvement to another PR and just steal from `LinAlg` for now.

Is this similar to what @dwincort  suggested in this [issue](https://github.com/bolt12/laop/issues/21)?

src/InductiveMatrix.hs
bolt12: I did but I get this error:

```haskell
The exact Name ‘k’ is not in scope
  Probable cause: you used a unique Template Haskell name (NameU), 
  perhaps via newName, but did not bind it
  If that's it, then -ddump-splices might be useful
```

conal:
> Is this similar to what @dwincort suggested in this [issue](https://github.com/bolt12/laop/issues/21)?

I don’t think so. More like his `Shape` GADT and `HasShape` class in PR \#32.

goldfirere:
Just catching up here.

In this particular case, it would seem to solve the problem to allow a `COMPLETE` pragma even when the result type is polymorphic. Is that correct? I don't know why we have the result-type-must-be-concrete restriction, but maybe lifting it is easier than allowing orphans.

A little more poking about reveals relevant thread https://ghc-devs.haskell.narkive.com/NXBBDXg1/suppressing-false-incomplete-pattern-matching-warnings-for-polymorphic-pattern-synonyms where I floated this same idea previously, to get shot down. But I'm still not sure exactly we can't have polymorphic complete sets. We allow polymorphic instances, after all. Sadly, that thread points out that order of patterns matters, making the general solution even harder to come by.

Also relevant is https://gitlab.haskell.org/ghc/ghc/-/issues/14422, which asks for orphan `COMPLETE` pragmas that are not universal.

I don't have any good grand ideas here. My intuition is that getting this right will require a good deal of both design and implementation. :(

sgraf812:
I have an idea how to tackle the problem in https://gitlab.haskell.org/ghc/ghc/-/issues/18277#note_287827. It just requires a bit of reshuffling how COMPLETE pragmas are serialised in interface files, but should be simpler to work on that what we currently have.

conal:
> In this particular case, it would seem to solve the problem to allow a `COMPLETE` pragma even when the result type is polymorphic. Is that correct?

Yes, and polymorphic `COMPLETE` pragmas (or GHC determining completeness on its own) would be ideal.

bolt12:
@conal I think everything is set now! I still could not get `deriving via`to work, but we can tackle that in the future!

conal:
Merged. Thank you @bolt12 for the care and persistence!

bolt12:
Hooray! Thanks @conal for all your help!

src/Category.hs
```diff
@@ -197,7 +196,10 @@ instance Category k => Category (ViaCartesian p k) where
   type Obj' (ViaCartesian p k) a = Obj k a
   id = ViaCartesian id
   ViaCartesian g . ViaCartesian f = ViaCartesian (g . f)
-deriving instance Monoidal  p k => Monoidal  p (ViaCartesian p k)
+instance Cartesian p k => Monoidal p (ViaCartesian p k) where
+  f ### g = (f . exl) &&& (g . exr)
+instance (Category k, Representable r, CartesianR r p (ViaCartesian p k)) => MonoidalR r p (ViaCartesian p k) where
+  rmap fs = fork (liftR2 (.) fs exs)
 deriving instance Cartesian p k => Cartesian p (ViaCartesian p k)
```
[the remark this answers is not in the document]
dwincort: There are actually two separate things going on here.  I can definitely fix one of them, but the other is a little more challenging.

---

First, there is the error from the line
```haskell
deriving via (ViaCartesian (:*:) (L s)) instance Monoidal (:*:) (L s)
```
that says "The exact Name ‘k’ is not in scope" and suggests a Template Haskell problem.  We have no TH here, so that's a bit of a red herring.  The real problem has to do with _implicit kinds_, which I figured out by finding this unscoped `k` in the generated code (`:set -ddump-deriv`).  The fix is surprisingly easy once the problem is identified: make the kind _explicit_!  To do this, we simply change the `newtype` definition to:
```haskell
newtype ViaCartesian (p :: k1 -> k1 -> k1) (k :: k1 -> k1 -> *) (a :: k1) (b :: k1) = ViaCartesian (k a b)
```
With that new definition, the `deriving via` line from `InductiveMatrix` works fine.  (Well, almost — you need t add `Semiring s` as a constraint.)

---

The second one is a little more challenging.  @conal is exactly right in his description of the problem, but the solution he suggested didn't seem to work for me.  When I tried to add the line `type role MonoidalR representational _ _`, I got an error saying this would only work if I enabled incoherent instances.  (I don't want to enable them, but even when I did, I just got the same error message over and over.)  Based on what I know about roles, this sort of makes sense to me.

Another "solution" is to let GHC know that every representable functor we might use does, indeed, have role representational.  We can do that by adding a constraint like:
```haskell
type Representational t = (forall a b. Coercible (t a) (t b))
```
and then adding `Representational r` where we need it.  I use quotes when I say this is a "solution" because I'm not convinced this constraint is an easy burden to satisfy.

This whole situation is a little frustrating.  We're using these newtype wrappers entirely to help with deriving.  However, now that they exist, someone could theoretically create a representable functor that uses a type family that matches on `ViaCartesian` specifically in order to distinguish whether we have a `k` or a `ViaCartesian p k`, which in turn means we can't use it for deriving.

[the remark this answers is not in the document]
goldfirere: I didn't follow all the details here, but I think you mean `Representational` to be

```hs
type Representational t = (forall a b. Coercible a b => Coercible (t a) (t b))
```

Note the antecedent. As written your `Representational` is really `Phantom`.

And if GHC is suggesting adding `-XIncoherentInstances` but then still being unhappy, that sounds like a bug.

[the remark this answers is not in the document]
dwincort: @goldfirere Thanks.  Yes, `Coercible a b` is definitely a necessary constraint in there!

Now that I play with it a bit, I'm wondering if this `Representational` constraint really _is_ that big of a burden.  I'm not sure.

Lastly, I also realized that I got the kind signature just slightly wrong.  It should be:
```haskell
newtype ViaCartesian (p :: k2 -> k1 -> k1) (k :: k1 -> k1 -> *) (a :: k1) (b :: k1) = ViaCartesian (k a b)
```
(Note the `k2` in `p`'s kind.)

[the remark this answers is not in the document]
conal: See PR #56.

PR #56, opened 2020-09-01
Deriving via no phantom

conal:
Following the conversation about wonky `deriving via` [here](https://github.com/conal/linalg/pull/54#discussion_r481393523), I was poking around and noticed that `ViaCartesian` and `ViaCocartesian` had phantom type parameters (product and coproduct respectively). On a lark, I tried removing those parameters, and the result compiled. The previously broken deriving-via `Monoidal` instance in `InductiveMatrix` then started working (after adding `Semiring s`).

The deriving-via `MonoidalR` instance still needs work, due to the representational equality issue also discussed [here](https://github.com/conal/linalg/pull/54#discussion_r481393523).

### DIFF
--- a/src/Category.hs
+++ b/src/Category.hs
@@ -249,46 +249,46 @@ class BicartesianR r p p k => BiproductR r p k
 
 -- | The 'ViaCartesian' type is designed to be used with `DerivingVia` to derive
 -- `Associative` and `Symmetric` instances using the `Cartesian` operations.
-newtype ViaCartesian p k a b = ViaCartesian (k a b)
-instance Category k => Category (ViaCartesian p k) where
-  type Obj' (ViaCartesian p k) a = Obj k a
+newtype ViaCartesian k a b = ViaCartesian (a `k` b)
+instance Category k => Category (ViaCartesian k) where
+  type Obj' (ViaCartesian k) a = Obj k a
   id = ViaCartesian id
   ViaCartesian g . ViaCartesian f = ViaCartesian (g . f)
-instance Cartesian p k => Monoidal p (ViaCartesian p k) where
+instance Cartesian p k => Monoidal p (ViaCartesian k) where
   f ### g = (f . exl) &&& (g . exr)
-instance (Representable r, CartesianR r p k) => MonoidalR r p (ViaCartesian p k) where
+instance (Representable r, CartesianR r p k) => MonoidalR r p (ViaCartesian k) where
   rmap fs = fork (liftR2 (.) fs exs)
-deriving instance Cartesian p k => Cartesian p (ViaCartesian p k)
-instance (CartesianR r p k, Representable r) => CartesianR r p (ViaCartesian p k) where
+deriving instance Cartesian p k => Cartesian p (ViaCartesian k)
+instance (CartesianR r p k, Representable r) => CartesianR r p (ViaCartesian k) where
   exs = ViaCartesian <$> exs
   dups = ViaCartesian dups
 
-instance Cartesian p k => Associative p (ViaCartesian p k) where
+instance Cartesian p k => Associative p (ViaCartesian k) where
   lassoc = second exl &&& (exr . exr)
   rassoc = (exl . exl) &&& first  exr
-instance Cartesian p k => Symmetric p (ViaCartesian p k) where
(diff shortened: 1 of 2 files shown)


src/InductiveMatrix.hs
```diff
@@ -24,7 +25,13 @@ import Category.Isomorphism
...
 infixr 2 :|#
 
-type V' a = (Representable a, Eq (Rep a))
+type Representational r =
+  ((forall p q. Coercible p q => Coercible (r p) (r q)) :: Constraint)
+
+class    (Representable r, Representational r) => Representable' r
+instance (Representable r, Representational r) => Representable' r
+
+type V' a = (Representable' a, Eq (Rep a))
```
dwincort, line 34: Ahh, I thought this might be what to do too, but it's not.  Things get really hairy if we require `Representational` as part of `V`.  Instead, just add `Representational r` to the contexts of the `JoinL` and `ForkL` constructors as well as in the instance contexts for `CartesianR`, `MonoidalR`, and `CocartesianR`.  At least for me, that passed the type checker.

src/Category.hs
```diff
@@ -249,46 +249,46 @@ class BicartesianR r p p k => BiproductR r p k
...
-instance Cocartesian co k => Monoidal co (ViaCocartesian co k) where
+instance Cocartesian co k => Monoidal co (ViaCocartesian k) where
   f ### g = (inl . f) ||| (inr . g)
-instance (CocartesianR r p k, Representable r) => MonoidalR r p (ViaCocartesian p k) where
+instance (CocartesianR r p k, Representable r) => MonoidalR r p (ViaCocartesian k) where
   rmap fs = join (liftR2 (.) ins fs)
-deriving instance Cocartesian co k => Cocartesian co (ViaCocartesian co k)
-instance (CocartesianR r p k, Representable r) => CocartesianR r p (ViaCocartesian p k) where
+deriving instance Cocartesian co k => Cocartesian co (ViaCocartesian k)
+instance (CocartesianR r p k, Representable r) => CocartesianR r p (ViaCocartesian k) where
```
dwincort, line 284: Actually, I used the `Representational r` trick here as well, allowing this instance to be derived.

conal:
Compiles now and looks pretty good to me.

src/InductiveMatrix.hs
conal: I found the same as well. Please let me know how it looks to you now.

src/Category.hs
conal: > Actually, I used the `Representational r` trick here as well, allowing this instance to be derived.

Fantastic\! Fixing now.

src/Category.hs
```diff
@@ -325,10 +334,10 @@ instance MonoidalClosed (:*) (->) (->) where
...
-deriving via (ViaCartesian   (:*) (->)) instance Symmetric   (:*) (->)
-deriving via (ViaCocartesian (:+) (->)) instance Associative (:+) (->)
-deriving via (ViaCocartesian (:+) (->)) instance Symmetric   (:+) (->)
+deriving via (ViaCartesian   (->)) instance Associative (:*) (->)
+deriving via (ViaCartesian   (->)) instance Symmetric   (:*) (->)
+deriving via (ViaCocartesian (->)) instance Associative (:+) (->)
+deriving via (ViaCocartesian (->)) instance Symmetric   (:+) (->)
 
 instance Representable r => MonoidalR r Ap (->) where
   rmap rab (Ap ra) = Ap (liftR2 ($) rab ra)
```
dwincort, line 343: Now that we support deriving `Monoidal` and `MonoidalR`, these instance can actually be derived too:
```haskell
deriving via (ViaCartesian   (->)) instance Monoidal    (:*) (->)
deriving via (ViaCocartesian (->)) instance Monoidal    (:+) (->)
deriving via (ViaCartesian   (->)) instance RepresentableR r => MonoidalR r Ap (->)
deriving via (ViaCocartesian (->)) instance RepresentableR r => MonoidalR r RepAnd (->)
```
The only instances we must write by hand for `->` are `Category`, `Cartesian` and `Cocartesian` (and their `R` counterparts), and `Closed` and `MonoidalClosed`.

PR #57, opened 2020-09-02
Consolidate & stylize pattern synonyms

conal:
Some shuffling to clarify the pattern synonym situation until it improves.


conal:
We could move these shared pattern definitions to a CPP include file.

bolt12:
> We could move these shared pattern definitions to a CPP include file.

If that solves the issue I think it's worth a try, since we need to wait for GHC to be improved in this regard

conal:
> If that solves the issue I think it's worth a try, since we need to wait for GHC to be improved in this regard

More a cosmetic convenience than a solution, since we still have multiple specialized pattern synonyms rather than versions that can be used polymorphically.

PR #59, opened 2020-09-02
Add FunctorMatrix

dwincort:
This module adds matrices of the form `g (f s)` where `g` and `f` are representable functors.

There's one weird part of the definition, which is that I defined the `Obj'` type as:
```haskell
type Obj' (L s) a = (Semiring s, V a)
```
If I didn't include `Semiring s` here, then I didn't have that information when I needed it in the `LinearMap` instance.  That said, by including it in `Obj'`, I no longer need to have it in the context of the `Category` instance itself.

### DIFF
--- a/src/Category/Isomorphism.hs
+++ b/src/Category/Isomorphism.hs
@@ -88,10 +88,8 @@ joinIso = join :<-> unjoin
 forkIso :: (CartesianR r p k, Obj2 k a b) => r (a `k` b) <-> (a `k` p r b)
 forkIso = fork :<-> unfork
 
-curryIso :: ((a :* b) -> c) <-> (a -> (b -> c))
+curryIso :: (MonoidalClosed p e k, Obj3 k a b c) => ((a `p` b) `k` c) <-> (a `k` (b `e` c))
 curryIso = curry :<-> uncurry
-
--- TODO: generalize curry from (->) to cartesian closed
 
 newIso :: Newtype a => a <-> O a
 newIso = unpack :<-> pack
--- a/src/Category/Isomorphism.hs
+++ b/src/Category/Isomorphism.hs
@@ -114,3 +114,6 @@ type f <--> g = forall a. f a <-> g a
 
 fmapIso :: Functor f => a <-> b -> f a <-> f b
 fmapIso (f :<-> g) = (fmap f :<-> fmap g)
+
+flipIso :: (a -> b -> c) <-> (b -> a -> c)
+flipIso = flip :<-> flip
--- a/src/Category/Isomorphism.hs
+++ b/src/Category/Isomorphism.hs
@@ -115,3 +115,10 @@ fmapIso (f :<-> g) = (fmap f :<-> fmap g)
 
 flipIso :: (a -> b -> c) <-> (b -> a -> c)
 flipIso = flip :<-> flip
+
+distributeIso :: (Representable f, Representable g) => f (g a) <-> g (f a)
+-- distributeIso = inv repIso . fmapIso (inv repIso) . flipIso . fmapIso repIso . repIso
+distributeIso = distribute :<-> distribute
+
+collectIso :: (Representable f, Representable g) => f a <-> b -> f (g a) <-> g b
+collectIso f = fmapIso f . distributeIso
--- a/src/FunctorMatrix.hs
+++ b/src/FunctorMatrix.hs
@@ -25,26 +25,22 @@ newtype L (s :: *) (f :: * -> *) (g :: * -> *) = L { unL :: g (f s) }
   deriving (Show, Generic)
 
 instance Newtype (L s f g)
+-- type O (L s f g) = g (f s)
+
 
 instance LinearMap L where
   mu = inv newIso . distributeIso . fmapIso dotIso . newIso
-  --                          L s f g
-  -- newIso               ==> g (f s)
-  -- fmapIso dotIso       ==> g (f s -> s)
-  -- repIso               ==> Rep g -> f s -> s
-  -- flipIso              ==> f s -> Rep g -> s
-  -- fmapIso (inv repIso) ==> f s -> g s
-  -- inv newIso           ==> F.L s f g
+  --                    L s f g
+  -- newIso         ==> g (f s)
+  -- fmapIso dotIso ==> g (f s -> s)
+  -- distributeIso  ==> f s -> g s
+  -- inv newIso     ==> F.L s f g
 
-  -- mu = fwd :<-> rev
-  --   where
-  --     fwd (L gfs) = F.L $ flip fmap gfs . dot
-  --     rev (F.L m) = L $ collect m $ unL id
 
 instance Category (L s) where
   type Obj' (L s) a = (Semiring s, V a)
   id = L basis
-  b . a
+  L b . L a
   -- Trying to prove:
   --  L b . L a = L $ fmap (flip fmap (distribute a) . dot) b
   -- which is synonymous with
@@ -93,8 +89,18 @@ instance Category (L s) where
   --     $ \x -> fmap (dot (fmap (dot x) (unL a))) (unL b)
   --
   -- pointfree
-   = L $ fmap dot' $ distribute
-       $ flip fmap (unL b) . dot . flip fmap (unL a) . dot
+   = L $ fmap dot' $ distribute (flip fmap b . dot . flip fmap a . dot)
+  --
+  --  ???
+   -- = L $ fmap dot' $ fmap (dot . flip fmap (distribute a) . dot) b
+  --
+  --  fuse fmap; dot' . dot = id
+   -- = L $ fmap (flip fmap (distribute a) . dot) b
+
+
+-- Need to prove:
+--     distribute (flip fmap b . dot . flip fmap a . dot)
+--  == fmap (dot . flip fmap (distribute a) . dot) b
 
 
 -- How can I bring dot and dot' together to form an identity?
@@ -102,7 +108,7 @@ instance Category (L s) where
 --
 -- Are either of these statements true?
 -- distribute (g . f) ?= fmap (. f) (distribute g)
--- distribute (\x -> fmap (g x) y) ?= fmap (distribute g) y
+-- distribute (flip fmap y . f) ?= fmap (distribute f) y
 
 
 
@@ -115,25 +121,25 @@ instance Category (L s) where
 
 instance (Representable f, Representable g, Semiring s) => Additive (L s f g) where
   zero = pureL zero
-  (+)  = pointwise2 (+)
+  (+)  = liftL2 (+)
 
 pureL :: (Representable f, Representable g) => a -> L a f g
 pureL = pack . pureRep . pureRep
 
-pointwise2 :: (Representable f, Representable g) => (a -> b -> c) -> L a f g -> L b f g -> L c f g
-pointwise2 = inNew2 . liftR2 . liftR2
+liftL2 :: (Representable f, Representable g) => (a -> b -> c) -> L a f g -> L b f g -> L c f g
+liftL2 = inNew2 . liftR2 . liftR2
 
 fork2LIso :: (L s a c :* L s a d) <-> L s a (c :*: d)
 fork2LIso = inv newIso . inv newIso . (newIso ### newIso)
 
-join2LIso :: (Representable a, Representable c, Representable d) => (L s c a :* L s d a) <-> L s (c :*: d) a
+join2LIso :: (C3 Representable a c d) => (L s c a :* L s d a) <-> L s (c :*: d) a
 join2LIso = inv newIso . distributeIso . inv newIso . (distributeIso . newIso ### distributeIso . newIso)
 
-forkLIso :: Representable r => r (L s a b) <-> L s a (r :.: b)
-forkLIso = inv newIso . inv newIso . fmapIso newIso
+forkLIso :: RepresentableR r => r (L s a b) <-> L s a (r :.: b)
+forkLIso = inv newIso . inv newIso . coerceIso
 
-joinLIso :: (Representable b, Representable r) => r (L s a b) <-> L s (r :.: a) b
-joinLIso = inv newIso . collectIso (inv newIso) . fmapIso newIso
+joinLIso :: (Representable b, RepresentableR r) => r (L s a b) <-> L s (r :.: a) b
+joinLIso = inv newIso . collectIso (inv newIso) . coerceIso
 
 instance Semiring s => Cartesian (:*:) (L s) where
   (&&&)   = curry $ isoFwd fork2LIso
--- a/src/FunctorMatrix.hs
+++ b/src/FunctorMatrix.hs
@@ -1,0 +1,94 @@
+{-# LANGUAGE UndecidableInstances #-}
+
+{- |
+"Functor Matrices", or a "matrix" represented by a representable functor of
+representable functors.
+-}
+
+module FunctorMatrix where
+
+import CatPrelude
+
+import qualified LinearFunction as F
+import LinearFunction hiding (L(..))
+import Category.Isomorphism
+
+-------------------------------------------------------------------------------
+-- | Representation and its denotation
+-------------------------------------------------------------------------------
+
+type V' a = (Foldable a, Representable a, Eq (Rep a))
+
+class    V' a => V a
+instance V' a => V a
+
+type VR r = (V r, Representational1 r)
+
+-- | Compositional linear map representation.
+data L s f g where
+  L :: (Representable f, Representable g) => { unL :: g (f s) } -> L s f g
+
+instance LinearMap L where
+  mu = fwd :<-> rev
+    where
+      fwd (L gfs) = F.L $ flip fmap gfs . dot
+      rev (F.L m) = L $ collect m $ unL id
+
+instance Category (L s) where
+  type Obj' (L s) a = (Semiring s, V a)
+  id = L basis
+  L b . L a = L $ flip cotraverse a . dot <$> b
+
+instance (Representable f, Representable g, Semiring s) => Additive (L s f g) where
+  zero = L $ pureRep (pureRep zero)
+  (+)  = pointwise (+)
+
+pointwise :: (Representable f, Representable g) => (a -> b -> c) -> L a f g -> L b f g -> L c f g
+pointwise f (L a) (L b) = L $ liftR2 (liftR2 f) a b
+
+
+instance Semiring s => Cartesian (:*:) (L s) where
+  L gfa &&& L g'fa = L $ gfa :*: g'fa
+  unfork2 (L (x :*: y)) = (L x, L y)
+
+pattern (:&) :: (Cartesian p k, Obj3 k a c d)
+             => (a `k` c) -> (a `k` d) -> (a `k` (c `p` d))
+pattern f :& g <- (unfork2 -> (f,g)) where (:&) = (&&&)
+{-# COMPLETE (:&) :: L #-}
+
+instance Semiring s => Cocartesian (:*:) (L s) where
+  L gfa ||| L gf'a = L $ liftR2 (:*:) gfa gf'a
+  unjoin2 (L l) = (L $ fstP <$> l, L $ sndP <$> l)
+    where
+      fstP (x :*: _) = x
+      sndP (_ :*: y) = y
+
+pattern (:|) :: (Cocartesian co k, Obj3 k a b c)
+             => (a `k` c) -> (b `k` c) -> ((a `co` b) `k` c)
+pattern f :| g <- (unjoin2 -> (f,g)) where (:|) = (|||)
+{-# COMPLETE (:|) :: L #-}
+
+instance (VR r, Semiring s) => CartesianR r (:.:) (L s) where
+  fork   = L . Comp1 . fmap unL
+  unfork = fmap L . unComp1 . unL
+
+pattern Fork :: (CartesianR h p k, Obj2 k f g) => h (k f g) -> k f (p h g)
+pattern Fork ms <- (unfork -> ms) where Fork = fork
+{-# COMPLETE Fork :: L #-}
+
+instance (VR r, Semiring s) => CocartesianR r (:.:) (L s) where
+  join   = L . fmap Comp1 . distribute . fmap unL
+  unjoin = fmap L . distribute . fmap unComp1 . unL
+
+pattern Join :: (CocartesianR h p k, Obj2 k f g) => h (k f g) -> k (p h f) g
+pattern Join ms <- (unjoin -> ms) where Join = join
+{-# COMPLETE Join :: L #-}
+
+
+instance Semiring s => Biproduct (:*:) (L s)
+instance (VR r, Semiring s) => BiproductR r (:.:) (L s)
+
+deriving via ViaCartesian (L s) instance Semiring s => Monoidal (:*:) (L s)
+
+deriving via ViaCartesian (L s)
+  instance (VR r, Semiring s) => MonoidalR r (:.:) (L s)
--- a/src/FunctorMatrix.hs
+++ b/src/FunctorMatrix.hs
@@ -9,7 +9,6 @@ module FunctorMatrix where
 
 import CatPrelude
 
-import qualified LinearFunction as F
 import LinearFunction hiding (L(..))
 import Category.Isomorphism
 
@@ -24,15 +23,26 @@ instance V' a => V a
 
 type VR r = (V r, Representational1 r)
 
--- | Compositional linear map representation.
-data L s f g where
-  L :: (Representable f, Representable g) => { unL :: g (f s) } -> L s f g
+-- | Generalized matrices.
+newtype L (s :: *) (f :: * -> *) (g :: * -> *) = L { unL :: g (f s) }
+  deriving (Show, Generic)
+
+instance Newtype (L s f g)
 
 instance LinearMap L where
-  mu = fwd :<-> rev
-    where
-      fwd (L gfs) = F.L $ flip fmap gfs . dot
-      rev (F.L m) = L $ collect m $ unL id
+  mu = inv newIso . fmapIso (inv repIso) . flipIso . repIso . fmapIso dotIso . newIso
+  --                          L s f g
+  -- newIso               ==> g (f s)
+  -- fmapIso dotIso       ==> g (f s -> s)
+  -- repIso               ==> Rep g -> f s -> s
+  -- flipIso              ==> f s -> Rep g -> s
+  -- fmapIso (inv repIso) ==> f s -> g s
+  -- inv newIso           ==> F.L s f g
+
+  -- mu = fwd :<-> rev
+  --   where
+  --     fwd (L gfs) = F.L $ flip fmap gfs . dot
+  --     rev (F.L m) = L $ collect m $ unL id
 
 instance Category (L s) where
   type Obj' (L s) a = (Semiring s, V a)
@@ -49,7 +59,7 @@ pointwise f (L a) (L b) = L $ liftR2 (liftR2 f) a b
 
 instance Semiring s => Cartesian (:*:) (L s) where
   L gfa &&& L g'fa = L $ gfa :*: g'fa
-  unfork2 (L (x :*: y)) = (L x, L y)
+  unfork2 (L (gfa :*: g'fa)) = (L gfa, L g'fa)
 
 pattern (:&) :: (Cartesian p k, Obj3 k a c d)
              => (a `k` c) -> (a `k` d) -> (a `k` (c `p` d))
@@ -77,8 +87,8 @@ pattern Fork ms <- (unfork -> ms) where Fork = fork
 {-# COMPLETE Fork :: L #-}
 
 instance (VR r, Semiring s) => CocartesianR r (:.:) (L s) where
-  join   = L . fmap Comp1 . distribute . fmap unL
-  unjoin = fmap L . distribute . fmap unComp1 . unL
+  join   = L . cotraverse Comp1 . fmap unL
+  unjoin = fmap L . collect unComp1 . unL
 
 pattern Join :: (CocartesianR h p k, Obj2 k f g) => h (k f g) -> k (p h f) g
 pattern Join ms <- (unjoin -> ms) where Join = join
--- a/src/FunctorMatrix.hs
+++ b/src/FunctorMatrix.hs
@@ -12,9 +12,6 @@ import CatPrelude
 import LinearFunction hiding (L(..))
 import Category.Isomorphism
 
--------------------------------------------------------------------------------
--- | Representation and its denotation
--------------------------------------------------------------------------------
 
 type V' a = (Foldable a, Representable a, Eq (Rep a))
 
@@ -30,7 +27,7 @@ newtype L (s :: *) (f :: * -> *) (g :: * -> *) = L { unL :: g (f s) }
 instance Newtype (L s f g)
 
 instance LinearMap L where
-  mu = inv newIso . fmapIso (inv repIso) . flipIso . repIso . fmapIso dotIso . newIso
+  mu = inv newIso . distributeIso . fmapIso dotIso . newIso
   --                          L s f g
   -- newIso               ==> g (f s)
   -- fmapIso dotIso       ==> g (f s -> s)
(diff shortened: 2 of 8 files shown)


src/FunctorMatrix.hs
```diff
@@ -0,0 +1,94 @@
...
+type V' a = (Foldable a, Representable a, Eq (Rep a))
+
+class    V' a => V a
+instance V' a => V a
+
+type VR r = (V r, Representational1 r)
+
+-- | Compositional linear map representation.
+data L s f g where
+  L :: (Representable f, Representable g) => { unL :: g (f s) } -> L s f g
```
conal, line 29: Why the constraints on `f` and `g` here rather than only in `Obj (L s)`?

src/FunctorMatrix.hs
```diff
@@ -0,0 +1,94 @@
...
+-------------------------------------------------------------------------------
+
+type V' a = (Foldable a, Representable a, Eq (Rep a))
+
+class    V' a => V a
+instance V' a => V a
+
+type VR r = (V r, Representational1 r)
+
+-- | Compositional linear map representation.
```
conal, line 27: In what sense “compositional”? It’s less compositional than `InductiveMatrix`. Maybe call it “generalized matrices” or “generalized non-inductive matrices”

src/FunctorMatrix.hs
```diff
@@ -0,0 +1,94 @@
...
+data L s f g where
+  L :: (Representable f, Representable g) => { unL :: g (f s) } -> L s f g
+
+instance LinearMap L where
+  mu = fwd :<-> rev
+    where
+      fwd (L gfs) = F.L $ flip fmap gfs . dot
+      rev (F.L m) = L $ collect m $ unL id
```
conal, line 35: Begs to be written via isomorphic building blocks instead of `(:<->)` (which is unsafe). Hopefully clearer, as well as necessarily a true isomorphism.

src/FunctorMatrix.hs
conal: I’m unused to thinking about constraints in this position. Does your formulation imply that every linear map is represented as a *function* from dictionary pairs to generalized matrices? If so, I think we’d lose all of the efficiency advantages of matrices (data, not functions) and gain none of the efficiency advantages of functions (ideal for things like `id`, `exl`, `inl`, etc, especially for high dimensions).

dwincort: No, these constraints were unnecessary, and I've now removed them.

src/FunctorMatrix.hs
dwincort: It wasn't — that line was brought in accidentally in a copy-paste from `InductiveMatrix`.

src/FunctorMatrix.hs
```diff
@@ -0,0 +1,104 @@
...
+type VR r = (V r, Representational1 r)
+
+-- | Generalized matrices.
+newtype L (s :: *) (f :: * -> *) (g :: * -> *) = L { unL :: g (f s) }
+  deriving (Show, Generic)
+
+instance Newtype (L s f g)
+
+instance LinearMap L where
+  mu = inv newIso . fmapIso (inv repIso) . flipIso . repIso . fmapIso dotIso . newIso
```
dwincort, line 33: > Begs to be written via isomorphic building blocks instead of `(:<->)` (which is unsafe). Hopefully clearer, as well as necessarily a true isomorphism.

Done.  That said, I needed to use `flipIso`, which I just defined (`flip :<-> flip`).  I really feel like this should come for free from somewhere (perhaps something to do with currying and swapping) but I couldn't get it to work.

src/FunctorMatrix.hs
dwincort: > Does your formulation imply that every linear map is represented as a _function_ from dictionary pairs to generalized matrices?

I'm pretty sure the answer is no.  If I take away the GADT syntax, this is equivalent to writing:
```haskell
data L s f g = (Representable f, Representable g) => L { unL :: g (f s) }
```
As far as I understand, it's more like saying every linear map is a _pair_ of dictionary pairs and generalized matrices. 
 But, as I said, you were right: these constraints aren't actually necessary here.  Removing them allows me to rewrite `L` as a newtype, which is better anyway.

src/FunctorMatrix.hs
conal: Cool.

> perhaps something to do with currying and swapping

I think `dom` is your friend:

``` haskell
dom :: (Closed e k, Obj3 k c a d) => (d `k` c) -> ((c `e` a) `k` (d `e` a))
```

src/Category/Isomorphism.hs
```diff
@@ -88,11 +88,9 @@ joinIso = join :<-> unjoin
 forkIso :: (CartesianR r p k, Obj2 k a b) => r (a `k` b) <-> (a `k` p r b)
 forkIso = fork :<-> unfork
 
-curryIso :: ((a :* b) -> c) <-> (a -> (b -> c))
+curryIso :: (MonoidalClosed p e k, Obj3 k a b c) => ((a `p` b) `k` c) <-> (a `k` (b `e` c))
```
conal, line 91: Nice.

src/FunctorMatrix.hs
conal: Oh, hm. Maybe `dom` apply when `k = (->)`, but not otherwise.

src/FunctorMatrix.hs
```diff
@@ -0,0 +1,104 @@
...
+  --                          L s f g
+  -- newIso               ==> g (f s)
+  -- fmapIso dotIso       ==> g (f s -> s)
+  -- repIso               ==> Rep g -> f s -> s
+  -- flipIso              ==> f s -> Rep g -> s
+  -- fmapIso (inv repIso) ==> f s -> g s
+  -- inv newIso           ==> F.L s f g
```
conal, line 40: These type chains really help me follow along. Thanks.

It’s also interesting to consider linearity, since `L` (via `newIso`) must be applied to *linear* functions, and `dot` generates a linear function. There’s also an important linearity property `flip` (used implicitly here) that’s worth meditating on. In Agda, we’ll get/need to make these things explicit, which I think will be illuminating, since linear algebra is about linearity.

src/FunctorMatrix.hs
```diff
@@ -0,0 +1,104 @@
...
+
+  -- mu = fwd :<-> rev
+  --   where
+  --     fwd (L gfs) = F.L $ flip fmap gfs . dot
+  --     rev (F.L m) = L $ collect m $ unL id
+
+instance Category (L s) where
+  type Obj' (L s) a = (Semiring s, V a)
+  id = L basis
+  L b . L a = L $ flip cotraverse a . dot <$> b
```
conal, line 50: It’s not clear to me that this `(.)` definition is correct, i.e., satisfies the required homomorphism property. In contrast, `(.)` on inductive matrices is a matter of applying laws of biproduct categories, as noted in the comments. Any thoughts on how you would prove correctness, or better how to *calculate* the implementation (or a different one) from the specification?

src/FunctorMatrix.hs
```diff
@@ -0,0 +1,104 @@
...
+
+instance (Representable f, Representable g, Semiring s) => Additive (L s f g) where
+  zero = L $ pureRep (pureRep zero)
+  (+)  = pointwise (+)
+
+pointwise :: (Representable f, Representable g) => (a -> b -> c) -> L a f g -> L b f g -> L c f g
+pointwise f (L a) (L b) = L $ liftR2 (liftR2 f) a b
```
conal, line 57: Note that zero is also defined pointwise, as is scaling, so “`pointwise2`” might be a more fitting name.

Patterns like `fmap . fmap`, `pure . pure`, `liftA2 . liftA2`, `foldMap . foldMap`, and `traverse . traverse` appear whenever we compose functors, applicatives, foldables, and traversables (but not monads). Indeed that code already lives in various instances for functor composition `(:.:)`. This observation hints at making `(:.:)` explicit, either by changing the `L` definition or adding a custom `Newtype` instance.

Also, there’s [an idiom](https://github.com/conal/concat/blob/e71fe7dd73492abdfd432743ce2c1b249429fa5e/classes/src/ConCat/Misc.hs#L342-L350) I use frequently with `Newtype` instances that allows

``` haskell
pointwise = inNew2 . liftR2 . liftR2
```

or if you make `(:.:)` explicit,

``` haskell
pointwise = inNew2 . liftR2
```

I refer to this style as “[semantic editor combinators](http://conal.net/blog/posts/semantic-editor-combinators)” (SECs) and first noticed it when working on [*Tangible Functional Programming*](http://conal.net/papers/Eros/).

src/FunctorMatrix.hs
```diff
@@ -0,0 +1,104 @@
...
+
+
+instance Semiring s => Cartesian (:*:) (L s) where
+  L gfa &&& L g'fa = L $ gfa :*: g'fa
+  unfork2 (L (gfa :*: g'fa)) = (L gfa, L g'fa)
```
conal, line 62: This instance also begs to be written as a single isomorphism out of isomorphic pieces. Note how the two definitions are almost syntactic inverses of each other (especially if you drop the dollar operator).

The cost is probably not be worth the reward for this one example, but might be overall. One of the isomorphic building blocks has type `(f :*: g) x <-> f x :* g x`. The others are `fork2Iso` and `newIso`.

Of course then we’d have to give `Cartesian` a single isomorphism-valued method rather than `(&&&)` and `unfork2`, so we could define them together.

Likewise for `Cocartesian`, `CartesianR`, and `CocartesianR`.

Food for thought.

src/FunctorMatrix.hs
```diff
@@ -0,0 +1,104 @@
...
+
+type V' a = (Foldable a, Representable a, Eq (Rep a))
+
+class    V' a => V a
+instance V' a => V a
+
+type VR r = (V r, Representational1 r)
+
+-- | Generalized matrices.
+newtype L (s :: *) (f :: * -> *) (g :: * -> *) = L { unL :: g (f s) }
```
conal, line 27: I think kind-annotating `s` suffices here.

src/FunctorMatrix.hs
dwincort: Yes!  This is something I definitely thought about when writing this out.  That is, using `:<->` may be unsafe, but composing these isomorphisms may also be incorrect if properties like linearity don't hold.

src/FunctorMatrix.hs
dwincort: That `inNew2` function is quite nice!  Yes, now that you mention it, using `:.:` explicitly could work, or similarly, deriving `Functor`, `Representable`, etc. for `L`.  I'll play around with it.

src/FunctorMatrix.hs
dwincort: I actually wrote all of these as isomorphisms first, only to realize that `join2iso` and the others (https://github.com/conal/linalg/blob/master/src/Category/Isomorphism.hs#L77-L89) are already defined.  So, I broke the isomorphisms up and used them directly in the instance definitions.

I like the idea of adding those isomorphisms into the instances directly so one can choose whether to define the isomorphism or the two constituent functions (and then have a minimal pragma).

src/FunctorMatrix.hs
dwincort: Actually, the only one necessary is `f` (there's still a generic kind if you only fix `s`), but it looked a little silly to me to annotate `f` and none of the others.

conal: Oho! I see what you mean.

src/FunctorMatrix.hs
conal: > I like the idea of adding those isomorphisms into the instances directly so one can choose whether to define the isomorphism or the two constituent functions (and then have a minimal pragma).

Thanks for the reaction. Something to play with in another PR at some point.

Composable isomorphism programming has been addictive for me. It’s also potentially very useful, since non-invertibility is the only known source of theoretically necessary heat generation ([Landauer’s principle](https://en.wikipedia.org/wiki/Landauer%27s_principle)), IIUC. A friend who’s into this sort of thing sent me [a brand new paper](https://arxiv.org/abs/2009.00448) today on an efficient physical implementation of reversible computing.

Before tinkering with isomorphisms for composable and type-safe data parallel computing, I hadn’t realized that most of the category language is invertible. (Ditto for linear.)

src/FunctorMatrix.hs
```diff
@@ -47,48 +44,126 @@ instance LinearMap L where
...
-  zero = L $ pureRep (pureRep zero)
-  (+)  = pointwise (+)
+  zero = pureL zero
+  (+)  = pointwise2 (+)
+
+pureL :: (Representable f, Representable g) => a -> L a f g
+pureL = pack . pureRep . pureRep
+
+pointwise2 :: (Representable f, Representable g) => (a -> b -> c) -> L a f g -> L b f g -> L c f g
+pointwise2 = inNew2 . liftR2 . liftR2
```
conal, line 124: Given "`pureL`", I'd rename this one to "`lift2L`". Better yet (I think), define (derive!) `Representable (L a f g)` (and likewise in `LinearFunction`) and use `pureRep` and `liftR2`.

src/FunctorMatrix.hs
```diff
@@ -47,48 +44,126 @@ instance LinearMap L where
...
+join2LIso :: (Representable a, Representable c, Representable d) => (L s c a :* L s d a) <-> L s (c :*: d) a
+join2LIso = inv newIso . distributeIso . inv newIso . (distributeIso . newIso ### distributeIso . newIso)
 
-pointwise :: (Representable f, Representable g) => (a -> b -> c) -> L a f g -> L b f g -> L c f g
-pointwise f (L a) (L b) = L $ liftR2 (liftR2 f) a b
+forkLIso :: Representable r => r (L s a b) <-> L s a (r :.: b)
+forkLIso = inv newIso . inv newIso . fmapIso newIso
 
+joinLIso :: (Representable b, Representable r) => r (L s a b) <-> L s (r :.: a) b
+joinLIso = inv newIso . collectIso (inv newIso) . fmapIso newIso
```
conal: Already in `Category.Isomorphism`, right?

conal: Oh, but then you want to use in defining `(&&&)` etc. Hm.

src/FunctorMatrix.hs
```diff
@@ -0,0 +1,179 @@
...
+  -- = L $ fmap dot' $ distribute
+  --     $ \x -> fmap (flip dot (fmap (flip dot x) (unL a))) (unL b)
+  --
+  -- flip dot == dot
+  -- = L $ fmap dot' $ distribute
+  --     $ \x -> fmap (dot (fmap (dot x) (unL a))) (unL b)
+  --
+  -- pointfree
+   = L $ fmap dot' $ distribute
+       $ flip fmap (unL b) . dot . flip fmap (unL a) . dot
```
dwincort, line 97: I got a little stuck with this proof of a definition for `(.)` and have now spent longer than I'd care to admit on it :P.

I feel like I should be able to bring the `dot'` and one of the `dot`s together to form the identity, thus simplifying further, but I don't know how to deal with the `distribute` that's in the way.  I've tried to come up with some identities for `distribute`, but none of them seem quite right.  Every time I try to get around the `distribute`, I ended up in a weird place: the definitions type check, but the call to `distribute` vanishes altogether.

I may spend some time on this later, but if anyone has some inspiration, I'd love to hear it.

src/FunctorMatrix.hs
dwincort: > Better yet (I think), define (derive!) `Representable (L a f g)` (and likewise in `LinearFunction`) and use `pureRep` and `liftR2`.

I would _love_ to, but to be `Representable`, I would need the `a` type parameter at the end, right?  This is a classic problem we've faced: a data type can't be both `Representable` in one parameter and `Category` in two others.

I suppose we could try to define something like:
```haskell
class FunctorCat f where
  fmapCat :: (x -> y) -> f x a b -> f y a b
```
and then similar `DistributiveCat` and `RepresentableCat` classes, but this seems a little out there.  (Or maybe `Functor2` given that there are 2 types after the functor type?)

src/FunctorMatrix.hs
dwincort: Right.  The ones in `Category.Isomorphism` are defined in terms of the `Cartesian` and `Cocartesian` (and `R` versions) instances.  Instead, I define `L` versions of these isomorphisms that specifically _don't_ rely on those classes.

src/FunctorMatrix.hs
conal: > to be `Representable`, I would need the `a` type parameter at the end, right?

Oh, yeah\! And for linear functions, I think we also need `a` to be a semiring (so we can sample on a basis).

conal: > I suppose we could try to define something like: *\[… `FunctorCat` …\]*

Interesting\! Maybe we could even use a single `Functor` class, though not the usual Haskell one. More generally, functors go from one category to another. Maybe we could go from `(->)` to *natural transformations* over `(->)` and to *natural transformations* over *natural transformations* over `(->)`, which I is perhaps isomorphic to your `fmapCat` type above.

I wonder if this idea could really work.

src/FunctorMatrix.hs
```diff
@@ -0,0 +1,104 @@
...
+class    V' a => V a
+instance V' a => V a
+
+type VR r = (V r, Representational1 r)
+
+-- | Generalized matrices.
+newtype L (s :: *) (f :: * -> *) (g :: * -> *) = L { unL :: g (f s) }
+  deriving (Show, Generic)
+
+instance Newtype (L s f g)
```
conal, line 27: What is `O (L s f g)`, as in the types of `pack` and `unpack`? Probably worth a comment here so we know how to read uses of `newIso`.

src/FunctorMatrix.hs
```diff
@@ -0,0 +1,104 @@
...
+             => (a `k` c) -> (a `k` d) -> (a `k` (c `p` d))
+pattern f :& g <- (unfork2 -> (f,g)) where (:&) = (&&&)
+{-# COMPLETE (:&) :: L #-}
+
+instance Semiring s => Cocartesian (:*:) (L s) where
+  L gfa ||| L gf'a = L $ liftR2 (:*:) gfa gf'a
+  unjoin2 (L l) = (L $ fstP <$> l, L $ sndP <$> l)
+    where
+      fstP (x :*: _) = x
+      sndP (_ :*: y) = y
```
conal: `exlF` and `exrF` in `Misc`.

conal: Intriguingly, this `unjoin2` definition looks a lot like the default in `Cocartesian`. Maybe defining `inl` and `inr` here would be simpler.

Alternatively, maybe define and use `unzipF` (in `Misc`), like list `unzip` but for functor products.

Likewise, `jam` might be simpler than `(|||)`, and then be sure you keep defaulting `(###)` via `ViaCartesian` rather than via `ViaCocartesian`.

I’m a little nervous about the `inl`, `inr`, and `jam` suggestions, because then I think you’d have to take care not to use `(|||)` or `unjoin2` in defining `(.)`, to avoid circularity.

src/FunctorMatrix.hs
```diff
@@ -0,0 +1,104 @@
...
+
+pattern (:|) :: (Cocartesian co k, Obj3 k a b c)
+             => (a `k` c) -> (b `k` c) -> ((a `co` b) `k` c)
+pattern f :| g <- (unjoin2 -> (f,g)) where (:|) = (|||)
+{-# COMPLETE (:|) :: L #-}
+
+instance (VR r, Semiring s) => CartesianR r (:.:) (L s) where
+  fork   = L . Comp1 . fmap unL
+  unfork = fmap L . unComp1 . unL
```
conal: Well, those definitions came out nicely, didn't they?

src/FunctorMatrix.hs
dwincort: I think you might need to refresh -- I committed some changes several hours ago where I rewrote these instance definitions using isomorphisms.  Once I started, I too found that composable isomorphism programming was a little addictive :).

conal: Sure enough. I might have been looking at a particular commit and forgotten that I’d already seen the iso versions.

The `inl`, `inr`, `jam` definitions might be worth trying for comparison.

src/FunctorMatrix.hs
dwincort: I realized that `flip` is just a special case of `distribute`, and `distribute :<-> distribute` is definitely an isomorphism (`distribute` has an identity law).  This let's me clean up this definition quite a bit:
```haskell
mu = inv newIso . distributeIso . fmapIso dotIso . newIso
```

conal: > I realized that `flip` is just a special case of `distribute`, and `distribute :<-> distribute` is definitely an isomorphism (`distribute` has an identity law). This lets me clean up this definition quite a bit:
> 
> ``` haskell
> mu = inv newIso . distributeIso . fmapIso dotIso . newIso
> ```

Much prettier\!

Note also that *every* use of `distribute` on representable functors (RFs) denotes `flip`, which implies the `distribute :<-> distribute` isomorphism (for RFs). In other words, `index` (the denotation function) is a `Distributive` homomorphism, as is `tabulate`, s `index`/`tabulate` is a `Distributive` isomorphism.

There’s still the linearity issue to address. I don’t doubt that it’s true, since I’ve done the proof in the past. I’m happy to wait and address in the Agda implementation.

conal: From reading the `Distributive` documentation, every distributive functor is representable, so my remarks above don’t just apply to instances of `Representable`. Is the difference between `Distributive` and `Representable` that the functor/function isomorphism must be computable for the latter?

conal: Simplifying the definition of `mu` as you’ve done here is a big deal, because implementation correctness is defined mainly in terms of `mu`. The simplicity of every implementation proof and calculation thus depends critically on the simplicity of the `mu` definition.

src/FunctorMatrix.hs
conal: Because we’re fortunate enough to have a (computably) *bijective* semantic function (`mu`), we can *mechanically* construct correct implementations of all operations. Let `mu` be the forward direction and `mu'` the reverse. For composition, the homomorphism equation is `mu (g . f) = mu g . mu f`, or equivalently `g . f = mu' (mu g . mu f)`, which is in the form of a definition of `(.)` for representations in terms of the semantic bijection and `(.)` on denotations. It’s probably worth hand-optimizing the RHS to something more efficient, because `mu'` is quite expensive in this case.

This recipe is valid for all linear map representations and really only requires that `mu'` is a right inverse for `mu`.

With or without an invertible semantic function, class laws all follow from the homomorphism properties and denotational equality.

conal: I have not done this calculation myself. Maybe it’s harder than I expected.

dwincort: > It’s probably worth hand-optimizing the RHS to something more efficient, because `rev` is quite expensive in this case.

Yes.  I started with `b . a = isoRev mu $ isoFwd mu b . isoFwd mu a` and have tried to simplify by hand from there.  I managed to simplify the definition quite a bit (as seen in the above code), but I'm still left with two calls to `dot` and one call to `dot'`.  There really _should_ be a way to combine the inner `dot` with the `dot'` — after all, they're functioning on the same representable functor `f` — but I can't figure it out.

The main next step I need to do is to pull the `dot` "outside" the `distribute`.  Based on some experimentation, I believe the best path forward is to prove
```haskell
    distribute (flip fmap (unL b) . dot . flip fmap (unL a) . dot)
 == fmap (dot . flip fmap (distribute (unL a)) . dot) (unL b)
```
but I don't know how to proceed.  In the comments, I propose:
```haskell
distribute (flip fmap y . f) ?= fmap (distribute f) y
```
I'm pretty confident this holds for _some_ cases of `f` and `y`, but I haven't nailed down which ones (although I think if I do, it will provide the path toward a proof).  Also, this only takes us halfway; I'm still left trying to prove:
```haskell
    dot . flip fmap (distribute (unL a)) . dot
 == distribute (dot . flip fmap (unL a) . dot)
```
Frankly, just by looking at this, it seems a little surprising to me that it would hold, but these two expressions do have the same type, and I really do think they have the same functionality.  I think to prove this I would need to make good use of `dot`'s particular type signature (and perhaps that it is commutative in its arguments).

conal: > I started with `b . a = isoRev mu $ isoFwd mu b . isoFwd mu a` and have tried to simplify by hand from there.

Oh\! You were already on this track.

bolt12: I tried going via Free Theorem of `distribute` but that does not help :( It feels so close yet so far!

conal: Here’s a thought on how to make this calculation easier.

In `cb . ba`, don’t pull `cb` apart. Instead, treat it only as its denotation `b s -o c s`. Transpose (`distribute`) `ba :: b (a s)` to `a (b s)`, map the `cb` denotation over the result to get `a (c s)`, and transpose the result.

How do we explain these transpositions? I think the answer has to do with the n-ary coproduct law `h . join fs  = join ((h .) <$> fs)`. (The corresponding binary coproduct law is `h . (f ||| g) = h . f ||| h . g`.) Applying this law requires an `unjoin`, `fmap`, and `join`. The role of this law is especially clear for inductive matrices, because the representation matches the vocabulary in which the laws are expressed.

If we switched to a column-major representation, I don’t think we would need either of the two transpositions, *but* then the semantic function itself would involve a transposition (unlike with row major).

One detail mismatch above is *applying* the `cb` denotation vs *composing* it. In the inductive representation, `(.)` always decomposes linear maps to linear maps, except for the single case of `Scale s' . Scale s`. This mismatch is reflected in another oddity. With the inductive representation, our representable functors always have `Par1` at the leaves, while in the semantic model and non-inductive version, they don’t. For this reason at least, I don’t think we will be able to define the reverse direction of `mu` for inductive matrices, and indeed that direction is currently missing.

conal: Here’s another thought on making this calculation easier, related to the noodlings above.

Structure the definition of `(.)` very like the definition for inductive matrices. Since we lack the inductive structure, recreate it from an algebra of shapes.

conal:
I had an inspiration about how simplify the task of calculating (correct) operations on the non-inductive, row-major, representable matrix representation.

The idea is to factor the denotation through the inductive form, i.e., convert from non-inductive to inductive (new) and then from inductive to linear functions (done). A homomorphic first step guarantees that the composition is homomorphic because (a) the second step is homomorphic and, (b) homomorphisms compose.

I don’t remember having used this sort of factoring in denotational design, but it sure seems like a powerful technique. Denotations are homomorphisms, and homomorphisms have compositional structure.

“If you can’t solve a problem, then there’s an easier problem you can solve: find it.” - George Polya

conal:
I think we’ll need some new tools that will turn out to be useful for other purposes as well.

  - Dual space construction for *all* linear map representations. Currently, `LinearFunction` has `toScalarIso :: ... => a s <-> L s a`, specific to its `L` representation (linear functions).
  - An isomorphism between `f :.: Par1` and `f`.

I think we’ll want new categorical vocabulary for both of these (linear) isomorphisms.

src/FunctorMatrix.hs
```diff
@@ -0,0 +1,185 @@
+{-# LANGUAGE UndecidableInstances #-}
+
+{- |
+"Functor Matrices", or a "matrix" represented by a representable functor of
+representable functors.
+-}
+
+module FunctorMatrix where
```
conal, line 8: This module is about a *row-major* representation of matrices, and we may also want row-minor, so maybe a module name something like “`RowMajor`” would suit better.

conal:
> The idea is to factor the denotation through the inductive form, i.e., convert from non-inductive to inductive (new) and then from inductive to linear functions (done).

Although I expect this factoring to work, I suspect there’s a simpler way forward: phrase the semantic function for row-major matrices in terms of the categorical operations, and then exploit general properties of those operations.

Start with `b (a s)` (row-major representation for `a s -o b s`). Use `scale`, `join`, `fork`:

``` haskell
(fmap . fmap) scale :: b (a s) -> b (a (Par1 s -o Par1 s))
fmap join :: b (a (Par1 s -o Par1 s)) -> b ((a :.: Par1) s -o Par1 s)
fork :: b ((a :.: Par1) s -o Par1 s) -> ((a :.: Par1) s -o (b :.: Par1) s)

   fork . fmap join . (fmap . fmap) scale
=  fork . fmap (join . fmap scale)
:: b (a s) -> ((a :.: Par1) s -o (b :.: Par1) s)
```

Note that this recipe is compositionally isomorphic, i.e., made up of isomorphic building blocks.

Now we have two choices:

  - Introduce a linear isomorphism of type `par1Iso :: Iso k (c :.: Par1) c` for all linear `k`. Wrap the scale/join/fork recipe above.
  - Leave it as is, an isomorphism between `b (a s)` and `(a :.: Par1) s -o (b :.: Par1) s`.

conal:
> I think we’ll want new categorical vocabulary for both of these (linear) isomorphisms.

Now I realize that we can *define* `toScalarIso` and its dual in terms of `scale`:

``` haskell
Par1 :: a s -> Par1 (a s)

join . fmap scale :: a s -> (a :.: Par1) s -o Par1 s
fork . fmap scale :: a s -> Par1 s -o (a :.: Par1) s
```

Then drop the `Par1` with the `Iso k (a :.: Par1) s a`.

bolt12:
I tinkered a little with the Iso, is this what you had in mind? I needed to add a new data type for this. I'm also not sure how to write this with `k` instead of `(->)`.

```haskell
data Iso2 k a s b = (:<->:) { isoFwd2 :: a s `k` b s, isoRev2 :: b s `k` a s }

par1Iso :: Functor a => Iso2 (->) (a :.: Par1) s a
par1Iso = (fmap outP . unComp1) :<->: (Comp1 . fmap inP)
  where
    inP :: s -> Par1 s
    inP s = Par1 s
    outP :: Par1 s -> s
    outP (Par1 s) = s
```

conal:
> I tinkered a little with the Iso, is this what you had in mind? *\[…\]*

We already have category-polymorphic isomorphisms in `Category.Isomorphism`:

``` haskell
data Iso k a b = (:<->) { isoFwd :: a `k` b, isoRev :: b `k` a }
```

So I’m thinking of adding a *method* (in a new or existing class):

``` haskell
par1Iso :: ... => Iso k (c :.: Par1) c
```

Then define for various representations. I don’t know if we can define `par1Iso` when `k` is inductive matrices, however.

Well, probably don’t wire in `(:.:)` here and instead do something similar to `MonoidalR` etc.

BTW, you can write your `par1Iso` above with `(.)` for `Iso (->)`.

bolt12:
I see what I did wrong!

```haskell
par1Iso :: Functor a => (a :.: Par1) <--> a
par1Iso = fmap unPar1 . unComp1 :<-> Comp1 . fmap Par1
```

This is a lot more cleaner!

conal:
Yes, it’s tidier this way, but it’s still using `(.)` on `(->)`, not on `Iso (->)`. There’s a prettier way\!

bolt12:
@conal I'm afraid I'm not understanding what you mean by "using `(.)` on `(->)`, not on `Iso (->)`". Are you saying that I can give a `Category k` constrain and use `(.)` from Category, if so I tried that and get a kind error.

conal:
> Do you mean I could write it like this, instead ?

> 
> 
> ``` haskell
> par1Iso = fmapIso (unPar1 :<-> Par1) . (unComp1 :<-> Comp1)
> ```

Yes\! And now you can use `newIso` from `Category.Isomorphism` to make it even prettier.

bolt12:
Awesome! This style of programming is very interesting :smile:

conal:
> This style of programming is very interesting :smile:

It sure is\! I’m glad you’re enjoying it, too.

Here’s the definition I had in mind, which I think you already get:

``` haskell
par1Iso :: Functor a => (a :.: Par1) <--> a
par1Iso = fmapIso newIso . newIso
```

Oh, wow. If we require the functor `a` to be preserve representational equality, we can write it much more tersely:

``` haskell
par1Iso :: Representational1 a => (a :.: Par1) <--> a
par1Iso = coerceIso
```

This definition is also much more efficient, since it avoids traversing `a` and runs in no time at all.

conal:
Recall [the remarks above](https://github.com/conal/linalg/pull/59#discussion_r482615788) about isomorphisms (reversible computing) eliminating the only known theoretical barrier to computational efficiency.

bolt12:
Cool! 

Now all that's left, in order to go via denotation factorization, is a way to get from the non-inductive structure to the inductive structure, right?

conal:
> Now all that’s left, in order to go via denotation factorization, is a way to get from the non-inductive structure to the inductive structure, right?

Well, the `par1Iso` definitions above are in `Iso (->)`. I think we’ll either need to (a) generalize them to `Iso k` for linear `k` or (b) decide to leave the `Par1`s in the denotation type. I don’t know how to define `par1Iso` for inductive linear maps.

Note that the `fork . fmap (join . fmap scale)` recipe converts non-inductive “matrices” to *all* linear map representations.

conal:
> I think we’ll either need to (a) generalize them to `Iso k` for linear `k` or (b) decide to leave the `Par1`s in the denotation type.

The (b) option:

``` haskell
fromRowMajor :: Linear l => b (a s) -> l s (a :.: Par1) (b :.: Par1)
fromRowMajor = fork . fmap (join . fmap scale)
```

These building blocks are all isomorphic, we can instead write

``` haskell
rowMajorIso :: Linear l => b (a s) <-> l s (a :.: Par1) (b :.: Par1)
rowMajorIso = forkIso . fmapIso (joinIso . fmapIso scaleIso)
```

This definition misses the *linearity* of the `rowMajorIso` isomorphism, so I think there’s still some room for interesting and useful improvements. We’d have to address the issue of linear maps *as* a vector space (really, left semimodule), which I expect to be fairly easy and quite useful.

The functors (category homomorphisms) in my denotational design examples change categories and preserve objects (domain and codomain types). In contrast, the Haskell prelude’s `Functor` class preserve the category (`(->)`) and change the objects (`fmap :: (a -> b) -> (f a -> f b)`). The general categorical notion of functor allows changing category *and* the objects. The `rowMajorIso` above with `a :.: Par1` and `b :.: Par1` fits this more general pattern.

conal:
About my earlier remark,

> I don’t know if we can define `par1Iso` when `k` is inductive matrices, however.

One idea: add some form of natural isomorphism constructor to non-inductive matrices:

``` haskell
data L :: * -> (* -> *) -> (* -> *) -> * where
  ...
  IsoL :: ... => Iso (F.L s) a b -> L s a b
```

where `NatX` is the category of natural transformations. Are such natural transformations always linear?

Or replace `IsoL` with one half of the isomorphism:

``` haskell
data L :: * -> (* -> *) -> (* -> *) -> * where
  ...
  F :: ... => F.L s a b -> L s a b
```

Alternatively, replace with a remapping of *indices*:

``` haskell
data L :: * -> (* -> *) -> (* -> *) -> * where
  ...
  Reindex :: ... => (Rep b -> Rep a) -> L s a b
```

I guessed at the contravariance here (`Rep b -> Rep a`), because we’re dealing with indices/arguments. Indeed, I’d expect `Reindex r` to denote the following linear function:

``` haskell
reindex :: (Rep b -> Rep a) -> F.L s a b
reindex r = L (\ a -> tabulate (index a . r))
```

This `Reindex` closely resembles the following function from [*Type Your Matrices for Great Good*](https://github.com/bolt12/tymfgg-pearl):

``` haskell
fromF :: (...) => (a -> b) -> Matrix e a b
```

Recall that TYM matrices are parametrized by the “logarithms” (indices) rather than by representable functors.

Interestingly, `fromF` is covariant while `reindex` is contravariant. If we require *bijective* reindexing, then there’s no substantive difference. Without that requirement, however, I’m guessing that `fromF` uses zero for non-surjectivity and addition for non-injectivity, which is very like generalizing convolution to a functor, applicative, and monad. See [*Generalized Convolution and Efficient Language Recognition*](http://conal.net/papers/convolution), section 11.

Even `Reindex` is much more expressive than we need, in that most of its uses can be written with the other inductive matrix constructors. For instance, swapping and reassociation of pairs are defined without `Reindex` and could easily be redefined with `Reindex`. The identity and zero linear transformations can also be written via `Reindex`. We would need to define binary and n-ary unjoin and unfork on `Reindex`, and we could optimize many other operations to work more efficiently on `Reindex`, including `exl`, `exr`, and `dup` (and their n-ary variants). For `inl`, `inr`, and `jam`, I think the covariant `fromF` would work better.

bolt12:
Interesting! I was thinking about that `reindex` approach as well! When you say that `Reindex` is much more expressive than we need do you mean that in a bad way?

conal:
> When you say that `Reindex` is much more expressive than we need do you mean that in a bad way?

Partly, yes. I like that it’s necessarily linear and that it may have optimization value. On the other hand, it doesn’t feel quite in the spirit of inductive matrices, which are rooted in the fundamental operations on coproduct categories. Maybe it more naturally belongs to a different representation; or perhaps it’s just on the way to something simpler and more fundamental.

dwincort:
> Why redefine `Sum` and `Product` instead of using the versions from `Data.Monoid`?

They have the wrong instances for `Semigroup` and `Monoid` — they're using `Num` while we want to rely on `Additive` and `Semigroup`.

conal:
Although this PR still looks like work-in-progress, would you like a merge so that it doesn't get further out of sync with the master branch?

dwincort:
> Although this PR still looks like work-in-progress, would you like a merge so that it doesn't get further out of sync with the master branch?

It's up to you.  I just rebased onto the recent changes from your other PR, and it wasn't too bad.  That said, I've had a bit of code creep in this PR (like with the minor changes to Misc and CatPrelude), so getting it merged makes some sense.  Obviously, we can continue to work out the details of `RowMajor` once it's merged.

PR #60, opened 2020-09-07
Tweak Scalable and LinearMap

conal:
A couple of small changes:

  - The inconsistency of `LinearMap` with other classes has been bugging me. Fixed here.
  - Generalize `unscale`, so scaling is a general isomorphism, just as (binary and n-ary) fork and join are.


dwincort (approved):
These look like good changes to me.

I noticed this when writing an instance of `LinearMap` that I really wanted to give a constraint on `s`, but that was impossible in the old version.  Now that `s` is mentioned in the instance head, it's easy to do.

PR #61, opened 2020-09-10
Make it compile

bolt12:
Just fixed the `LinearMap` instance in the `InductiveMatrix` module


conal:
Thanks! I thought I'd already committed & pushed that change, but apparently not.

Issue #35, opened 2020-08-01
Definition of transposition


snowleopard:
I've seen transposition mentioned here a few times. Do we have a definition of what it means for linear maps?

While looking at #34, I remembered one curious definition that I learned from @simonpj.

Let `m : X ⊸ Y` be a linear map, `m @ x` stand for application of `m` to a vector `x`, and `x • y` stand for dot product of vectors `x` and `y`. Then, `transpose : (X ⊸ Y) -> (Y ⊸ X)` can be _uniquely_ defined by the following property:

```
∀ x : X, y : Y. (transpose m @ y) • x = y • (m @ x)
```

While this definition might seem somewhat mysterious, I like it because it doesn't involve any dual vector spaces, and can be used to mechanically prove that the natural implementation of `transpose` is correct:

```haskell
transpose :: L c r s -> L r c s
transpose (Scale s)  = Scale s
transpose (f :|# g)  = tr f :&# tr g
transpose (f :&# g)  = tr f :|# tr g
transpose (ForkL ms) = JoinL (fmap transpose ms)
transpose (JoinL ms) = ForkL (fmap transpose ms)
```

For example, here is a short proof that `transpose (Scale s) = Scale s` satisfies the above definition:

```
(transpose (Scale s) @ y) • x = (Scale s @ y) • x = y • (Scale s @ x)
```

The second step is a property of dot product and scalar multiplication.

By the way, it seems that there is currently no definition of `@`, i.e. the meaning of a linear map, in the repository. One might say, that composition `.@` does in fact define this meaning (since application and composition of linear maps are pretty much the same thing) but it may still be useful to have `@` written down explicitly, for example, to make it easier to prove that all other equations in the above implementation of `transpose` match the definition.

bolt12:
Oh that's interesting! Do you have any reference to that property and to why does it _uniquely_ defines `transpose`? 

I also think that a lot of discussions around `transposition` and other operations revolve more around their meaning in terms of the linear map denotation and so, denotationally, the meaning of linear map transposition is the one expressed in my code (#34 ) (I think) and from that we just need/can go back to its matrix representation which is what I'm trying to do in `trT2` at #34. So `trT` has the correct type signature wrt to the denotation and denotes the essence of the meaning of linear map transposition and `trT2` tries to go from that to a more familiar type signature. I hope what I'm saying makes sense, I'm still new to the whole "denotational design" headspace. 

I'm also trying to implement `@` and am running into a wall:

```haskell
[...]

PR #62, opened 2020-10-22
Transpose 2

bolt12:
Gave this an attempt and think I got it right (hopefully!)

Tagging #34 for keeping track of previous work.

### DIFF
--- a/src/LinearFunction.hs
+++ b/src/LinearFunction.hs
@@ -120,6 +120,12 @@ dotIso = dot :<-> dot'
 
 -- TODO: basis vs oneHot
 
+transpose :: (Eq (Rep a), Semiring s, Representable a, Representable b, Foldable b) => L s a b -> L s b a
+transpose = L . (\g -> dot' . g . dot) . tr . unL
+  where
+    tr :: (a k -> b k) -> ((b k) -> k) -> ((a k) -> k)
+    tr f = \x -> x . f
+
 -- Convert vector to vector-to-scalar linear map ("dual vector")
 toScalar :: (Representable a, Foldable a, Semiring s) => a s -> L s a Par1
 toScalar a = L (Par1 . dot a)
--- a/src/LinearFunction.hs
+++ b/src/LinearFunction.hs
@@ -120,6 +120,22 @@ dotIso = dot :<-> dot'
 
 -- TODO: basis vs oneHot
 
+{-
+`tr` is a sensible implementation and can be proven to be correct by using `(@@)` given the following
+reasoning:
+
+Let `m : X ⊸ Y` be a linear map, `m @ x` stand for application of `m` to a vector `x`,
+and `x • y` stand for dot product of vectors `x` and `y`.
+Then, `transpose : (X ⊸ Y) -> (Y ⊸ X)` can be uniquely defined by the following property:
+
+  `∀ x : X, y : Y. (transpose m @ y) • x = y • (m @ x)`
+  `∀ x : X, y : Y. dot (transpose m @ y) x = dot y (m @ x)`
+  `∀ x : X, y : Y. dot (transpose m @ y) x = dot y (m @ x)`
+
+However this approach is not consistent with the denotation and there is still some discussion about the
+type of `(@@)`.
+
+-}
 transpose :: (Eq (Rep a), Semiring s, Representable a, Representable b, Foldable b) => L s a b -> L s b a
 transpose = L . (\g -> dot' . g . dot) . tr . unL
   where
--- a/src/LinearFunction.hs
+++ b/src/LinearFunction.hs
@@ -121,23 +121,43 @@ dotIso = dot :<-> dot'
 -- TODO: basis vs oneHot
 
 {-
-`tr` is a sensible implementation and can be proven to be correct by using `(@@)` given the following
-reasoning:
+`transpose` is a sensible implementation and is inspired by the following reasoning:
 
 Let `m : X ⊸ Y` be a linear map, `m @ x` stand for application of `m` to a vector `x`,
 and `x • y` stand for dot product of vectors `x` and `y`.
 Then, `transpose : (X ⊸ Y) -> (Y ⊸ X)` can be uniquely defined by the following property:
 
   `∀ x : X, y : Y. (transpose m @ y) • x = y • (m @ x)`
-  `∀ x : X, y : Y. dot (transpose m @ y) x = dot y (m @ x)`
-  `∀ x : X, y : Y. dot (transpose m @ y) x = dot y (m @ x)`
 
-However this approach is not consistent with the denotation and there is still some discussion about the
-type of `(@@)`.
+So, if:
+  `transpose = L . (\g -> dot' . g . dot) . tr . unL`
+
+and:
+  (L l) @ x = l x
+
+then (handling the newtype wrapping/unwrapping implicitly):
+
+  ```
+  ∀ x : X, y : Y. (transpose m @ y) • x = y • (m @ x)
+
+  dot (transpose m @ y) x = dot y (m @ x)
+    === < def. transpose >
+  dot ((dot' . tr m . dot) @ y) x
+    === < function application >
+  dot (dot' (tr m (dot y))) x
+    === < point-free (get rid of x) >
+  dot (dot' (tr m (dot y)))
+    <=== < dot . dot' = id >
+  tr m (dot y)
+    <=== < def. tr >
+  (dot y) . m
+    <=== < introduce x; function application >
+  dot y (m @ x)
+  ```
 
 -}
 transpose :: (Eq (Rep a), Semiring s, Representable a, Representable b, Foldable b) => L s a b -> L s b a
-transpose = L . (\g -> dot' . g . dot) . tr . unL
+transpose (L m) = L $ dot' . tr m . dot
   where
     tr :: (a k -> b k) -> ((b k) -> k) -> ((a k) -> k)
     tr f = \x -> x . f
--- a/src/Category/Isomorphism.hs
+++ b/src/Category/Isomorphism.hs
@@ -1,22 +1,24 @@
 -- {-# OPTIONS_GHC -Wno-unused-imports #-} -- TEMP
 
 -- | Isomorphisms
-
 module Category.Isomorphism where
 
-import qualified GHC.Generics as G
-import GHC.Exts (Coercible,coerce)
 import Control.Newtype.Generics
+import GHC.Exts (Coercible, coerce)
+import qualified GHC.Generics as G
 
 import CatPrelude
 
 -------------------------------------------------------------------------------
+
 -- | Representation
+
 -------------------------------------------------------------------------------
 
 infix 0 :<->
+
 -- | Isomorphism
-data Iso k a b = (:<->) { isoFwd :: a `k` b, isoRev :: b `k` a }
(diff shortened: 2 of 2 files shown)


src/LinearFunction.hs
```diff
@@ -120,6 +120,47 @@ dotIso = dot :<-> dot'
...
 -- TODO: basis vs oneHot
 
+{-
+`transpose` is a sensible implementation and is inspired by the following reasoning:
+
+Let `m : X ⊸ Y` be a linear map, `m @ x` stand for application of `m` to a vector `x`,
+and `x • y` stand for dot product of vectors `x` and `y`.
+Then, `transpose : (X ⊸ Y) -> (Y ⊸ X)` can be uniquely defined by the following property:
+
+  `∀ x : X, y : Y. (transpose m @ y) • x = y • (m @ x)`
```
conal, line 130: I don't find this specification compelling.

bolt12: Below?

src/LinearFunction.hs
```diff
@@ -120,6 +120,47 @@ dotIso = dot :<-> dot'
...
+    <=== < introduce x; function application >
+  dot y (m @ x)
+  ```
+
+-}
+transpose :: (Eq (Rep a), Semiring s, Representable a, Representable b, Foldable b) => L s a b -> L s b a
+transpose (L m) = L $ dot' . tr m . dot
+  where
+    tr :: (a k -> b k) -> ((b k) -> k) -> ((a k) -> k)
+    tr f = \x -> x . f
```
conal, line 162: To me, this one line is the heart of a compelling specification. It’s fundamentally about duality and fits a very general pattern that goes back to Cayley and then in more modern form in continuations and the Yoneda embedding. Specifically, in `L s b a`, `b` and `a` denote not vectors, but rather `L s b Par1` and `L s a Par1` respectively, i.e., a linear map between “dual vectors” / “covectors” / “linear functionals”. I think the same is true for gradient “vectors” as computed by automatic differentiation. They don’t denote vectors, but rather linear functionals. Instead of “multiplying” them by a scalar (e.g. by a learning rate in deep learning), they get *applied* to the scalar. Although it appears to be awkward to make linear maps (such as these linear functionals) be scalars in Haskell, I think it will be simple and natural in Agda.

src/LinearFunction.hs
conal: Oops. I had added some comments but I think I accidentally hit "start a review" instead of "add single comment", so I had an incomplete review going. Fixed now.

src/LinearFunction.hs
conal, line 162: To me, this one line is the heart of a compelling specification. It’s fundamentally about duality and fits a very general pattern that goes back to Cayley and then in more modern form in continuations and the Yoneda embedding. Specifically, in `L s b a`, `b` and `a` denote not vectors, but rather `L s b Par1` and `L s a Par1` respectively, i.e., a linear map between “dual vectors” / “covectors” / “linear functionals”. I think the same is true for gradient “vectors” as computed by automatic differentiation. They don’t denote vectors, but rather linear functionals. Instead of “multiplying” them by a scalar (e.g. by a learning rate in deep learning), they get *applied* to the scalar. Although it appears to be awkward to make linear maps (such as these linear functionals) be scalars in Haskell, I think it will be simple and natural in Agda.

bolt12: I'm glad you find this compelling! I used the property above to see and prove that if my implementation satisfied it! After my current thoughts on "transpose" I decided to give this a try and I'm happy it worked out!

conal: BTW, I hope it was clear that the “one line” I was referring to is “`tr f = \ x -> x . f`”. I like distilling out this simpler form, independent from the `dot`/`dot'` isomorphism that introduces a different representation for the linear functionals. I can imagine a variety of `transpose`-like functions, each combining `tr` with an isomorphism with a different representation of linear functionals.

Also, aren’t we completely missing the crucial role linearity here? How do we know that `tr m` is linear or that `dot' . tr m . dot` is linear? As I understand it, linearity is exactly what motivates the choice of `k`, making `tr` an *isomorphism* and `transform` an *involution*. Without linearity (or some other constraint on functions), I think we’d need to universally quantify over all return types in order to get an isomorphism. But then most of those return types would lack any notion of linearity.

Since linearity is what linear algebra is about, I don’t know how to capture the essence of what we really mean in Haskell.

conal: I guess the heart of my dissatisfaction with the `transpose` specification is that it fails to distinguish (a) the essential `tr` from (b) the inessential choice of alternative representation and its accompanying isomorphism. Maybe it would be illuminating to factor the specification, implementation, and *proof* of `transpose` into *four* pieces:

1.  A version of `tr` with some of the `->` types replaced by *linear* functions and `k` generalized to any linear space over `k`. Rename this `tr` to something like “`cont`”.
2.  A specialization of this generalized `tr` with the linear space being scalars, so that we have linear functionals.
3.  For an *arbitrary* alternative representation of the linear functionals and an isomorphism between linear functionals and that type, a generalization of `transpose` that works on that alternative representation.
4.  Specialization of this generalized `transpose` to a particular representation and isomorphism, namely vectors and `dot`/`dot'`.

conal: Tagging the closely related conversation #35.

bolt12: Sounds amazing! It would be very cool if we indeed could generalize `tr` (more correctly `cont`) in Agda!

I also think that your concerns are valid!

> Without linearity (or some other constraint on functions), I think we’d need to universally quantify over all return types in order to get an isomorphism. 

I wonder if linearity _is the_ condition needed. I tried applying the same idea but for simple functions and failed. I don't think that the reason why I couldn't was because it lacks the linearity property, see for instance:

```haskell
f :: (a -> b) -> ((b -> Bool) -> (a -> Bool))
f x = \y -> y . x

d :: a -> (a -> Bool)
d a = const True

d' :: (a -> Bool) -> a
d' x = _  -- This isn't possible

fTranspose :: (a -> b) -> (b -> a)
fTranspose x = d' . f x . d
```

conal: I like this thought experiment\! I think you’re pointing out that if we (a) leave functions completely unconstrained and (b) choose `Bool` for the “continuation’s” return type `r`, then we don’t get an isomorphism. Some questions remain:

  - If we don’t constrain functions, is there a choice other than `r = Bool` that does work (i.e., lead to an isomorphism)?
      - In *classical* logic, I think `r = Void` works, but the proof is nonconstructive, so we don’t get computation. Languages like Agda relate computation instead to *intuitionistic* logic, in which `¬ (¬ a)` does not imply `a` (though we do have the reverse), so the isomorphism fails. In Haskell I think this failure looks like the absence of a function of type `(a -> Void) -> Void` to `a` (though we do have the reverse). If instead, we *universally quantify* over `r`, then we get the isomorphism back.
      - Does any other choice of `r` work? If not, can we prove that there is no such `r`?
  - For function constraints *other than linearity* (e.g., being affine, polynomial, or rational, or more simply a monoid homomorphism), what choices of `r` give us an isomorphism?

conal: I suspect that the biproduct categories play in important role here. In that setting (including linear algebra), the `dot`/`dot'` isomorphism of linear algebra follows from a simpler isomorphism, which is that `(s ⊸ s) ≅ s` where `s` is scalars. The reason is that `(A × B ⊸ C) ≅ (A ⊸ C) × (B ⊸ C)` in biproduct categories, and we can take `C = s`. Repeatedly applying this law gives a product of `s ⊸ s`. If we look beyond linear algebra to other biproduct categories, I think it still makes sense to say `(f s ⊸ s) ≅ f (s ⊸ s)`, where `f` is a representable functor (so we can reasonably call `f s` a “vector”), which is part of the way to the linear algebra story. Maybe what sets linear algebra apart from other biproduct categories is mainly the isomorphism `(s ⊸ s) ≅ s`. For biproduct categories other than linear algebra, we can still talk about “matrices” whose elements are whatever isomorphic representation of `s ⊸ s` is available.

bolt12: > Some questions remain:

Pertinent questions and interesting remarks about classical logic! Biproduct categories might really play an important role! After reading your insights and trying to understand what makes it work in the case of linear transformations, I think that there might be a necessary property in order to have the needed isomorphism! I think that that property is by look a lot like something like this `(A × B ⊸ C) ≅ (A ⊸ C) × (B ⊸ C)` this is all just intuition and there might be a couple of missing pieces but this is all very exciting!

conal: Note that (a) in all cartesian categories we have `(A ⇒ C × D) ≅ (A ⇒ C) × (A ⇒ D)`, and in all cocartesian categories we have `(A + B ⊸ C) ≅ (A ⊸ C) × (B ⊸ C)`. In any bicartesian category, we can therefore decompose an arrow from coproducts to products into a “matrix” of non-coproducts to non-products. Specifically in biproduct categories, however, we have `(A × B ⊸ C) ≅ (A ⊸ C) × (B ⊸ C)`, because products and coproducts are the same (or isomorphic). As in your “Type Your Matrices” paper (Figure 1.24), the equivalence of products & coproducts lets us feed a product-producer (a fork) into a coproduct-consumer (a join)

bolt12: Also interesting enough the [wikipedia page for dual spaces](https://en.wikipedia.org/wiki/Dual_space#Transpose_of_a_linear_map) defines the transpose in the same way as `tr`. And it seems that this "Dual space" notion is only specific to vector spaces. I wonder if there's any work around this topic that aims to generalize this "Dual space" notion. I see references to CT and Adjunctions in the same Wikipedia article maybe there's something there!

Also interesting is that I found no reference to the transpose of a linear transformation being a linear relation in the same way the transpose of a function is a relation. But I guess I couldn't find that remark either in more popular literature.

bolt12: There's another interesting connection to this, regarding **weakest pre-conditions**. I talked with Prof. @JoseNunoOliveira about this transpose and dual space notions and showed him this thread as well. He thought it was very interesting and questioned about the relation with weakest pre-conditions. He argued that the function

```haskell
tr :: (a -> b) -> ((b -> Bool) -> (a -> Bool))
tr x = \y -> y . x
```

can be seen as the weakest pre-condition "equation". This is, the weakest constraint on inputs for outputs by `x` to fall within `y`.

In Haskell, for example:

```haskell
-- pre-condition type
type PreCondition a  = (a -> Bool)
-- pos-condition type
type PostCondition a = (a -> Bool)

-- weakest pre-condition
wp :: (a -> b) -> PostCondition b -> PreCondition a
wp f q = q . f
```

The special situation of a weakest precondition is nicely captured by the universal property:
```
f · Φp = Φq · f
 ≡
p = q · f
```

where `Φp :: a -> a` is the coreflexive relation which represents predicate `p :: (a -> Bool)` as a binary relation, `y Φp x ⇔ y = x ∧ p x`. In Haskell is not possible to have a total function that expresses this relation unless the data types are finite (see [here](https://github.com/bolt12/laop/blob/master/src/LAoP/Matrix/Internal.hs#L774))

More info about this weakest pre-condition calculus can be seen [here](http://www4.di.uminho.pt/~jno/ps/pdbc.pdf) (pag. 249)

bolt12: So building the correflexive `Φ :: (a -> Bool) -> a -> a` is the second half of the isomorphism that we cannot compute!

